PR #47631, opened 2016-07-29
X-Forwarded-Protocol should actually be X-Forwarded-Proto per RFC7239

charliewolf:
So I wasted a few hours debugging why none of my apps that do https redirects were respecting this header... turns out this is the culprit... Per [RFC7239](https://tools.ietf.org/html/rfc7239) and most common usage (nginx/flask/apache/etc), the header should be `X-Forwarded-Proto`.

We have worked around this in our app code for now but it is less than ideal.

### DIFF
--- a/src/proxy/engine.cpp
+++ b/src/proxy/engine.cpp
@@ -416,7 +416,7 @@ public:
 
 	bool isXForwardedProtocolTls(const HttpHeaders &headers)
 	{
-		QByteArray xfp = headers.get("X-Forwarded-Protocol");
+		QByteArray xfp = headers.get("X-Forwarded-Proto");
 		return (!xfp.isEmpty() && (xfp == "https" || xfp == "wss"));
 	}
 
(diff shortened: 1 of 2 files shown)


jkarneges:
Thanks for pointing this out! We don't want to break anyone relying on the old header (yet), so I've made it instead support both headers at the same time: ca7c5405ee419665232795aee344801d81b871da

charliewolf:
Awesome, thanks. Any idea when this will be deployed to fanout cloud?

jkarneges:
Just went live.

charliewolf:
Wow, you're efficient. Did you cut a release tag for it or is it just in master?

jkarneges:
We simply set Pushpin to version ca7c5405ee419665232795aee344801d81b871da in our deployment system. We often run development versions provided they are known to be stable and/or haven't deviated too far off from master.

charliewolf:
@jkarneges Would it be possible for you to cut a release? That way we can update our local dev environments (which use RPM or pacman packages)

charliewolf:
@jkarneges not sure if related to ca7c540 or not but I've recently noticed http requests from fanout cloud are coming without `X-Forwarded-Proto` set.

jkarneges:
Yes, it's only set if SSL is used and omitted otherwise.

PR #47674, opened 2019-07-03
support json payload on the input PULL/SUB sockets

jannic:
This version of the patch supports publishing a single item in json format on the pull/pub sockets.

The HTTP POST to /publish, the JSON structure is extended by an array items, to allow for publishing multiple items at once. Should that be possible on the zmq inputs, as well?

### DIFF
--- a/src/handler/engine.cpp
+++ b/src/handler/engine.cpp
@@ -2254,6 +2254,55 @@ private slots:
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


jkarneges:
The HTTP endpoint takes an array to enable high throughput, since it's a round trip operation. The PULL/SUB sockets don't have acknowledgements so this is not needed.

PR #47707, opened 2022-04-21
Missing IPv6 support -> Improve port parsing in pushpin config + Listen address building for condure

curdbecker:
Leverage QUrl for parsing and building host and port combinations. This PR also retains the original parsing approach, i.e. invalid host addresses are still treated as a 0.0.0.0 listen address for Condure and invalid ports are also rejected by QUrl identically. 

Primarily, this directly adds support for IPv6 addresses without adding further custom parsing. 

Otherwise, IPv6 listen addresses are not understood by Pushpin - which is quite frustrating, because Condure and all other components seem to support IPv6 perfectly, one just can't specify IPv6 listen addresses in Pushpin's config file.

### DIFF
--- a/src/runner/app.cpp
+++ b/src/runner/app.cpp
@@ -34,6 +34,7 @@
 #include <QFile>
 #include <QFileInfo>
 #include <QDir>
+#include <QUrl>
 #include "processquit.h"
 #include "log.h"
 #include "settings.h"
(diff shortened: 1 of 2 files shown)


jkarneges:
Thank you!

cb5890:
Thank you. That was fast! :)

Do you have any idea when you're going to release a new version? I don't mean to rush you or anything - we're just trying to find out whether it makes sense for us to set up an own container build environment for the time being, since we're planning on using the IPv6 support soon.

jkarneges:
Might be a couple of weeks before the next release.

cb5890:
Ok, thanks! 👍

PR #47726, opened 2023-08-31
move simplelogger into lib

jkarneges:


sima-fastly (approved):
thanks for this update. it will be useful for Runner updates coming up

PR #47728, opened 2023-08-31
remove webpki

jkarneges:
This fixes cargo audit. Required updating rustls.

The security issue in the webpki crate was recently fixed, but the fixed version is incompatible with the version of rustls we are using, and in the process of updating rustls it turns out we don't need to use webpki directly anymore.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -954,7 +954,6 @@ dependencies = [
  "thiserror",
  "time",
  "url",
- "webpki",
  "zmq",
 ]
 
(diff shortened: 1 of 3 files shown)


jkarneges:
This update only affects pushpin-publish. TLS still works as expected:

```
% target/debug/pushpin-publish -u $REALM_ID:$REALM_KEY --spec https://api.fanout.io/realm/$REALM_ID test yo
Published
```

sima-fastly (approved):
thanks for this update.
assuming tests are passing on test cache nodes, all looks great.

PR #47739, opened 2023-09-19
connection: ability to dynamically resize write buffers

jkarneges:
This change enables connections to support higher throughput by dynamically expanding the capacity of their write buffers. Capacity expands by a multiple of the configured buffer size ("blocks"). For example, if the buffer size is 1024, then when a write buffer is full it can expand to 2048 bytes, then 3072 bytes, and so on.

Size limits can be set for each connection as well as shared across connections. From the perspective of an individual connection, we refer to these limits as "blocks max" and "blocks available". Connections must have a minimum of 2 blocks (1 for the read buffer and 1 for the write buffer), and any additional blocks can go towards expanding the write buffer.

### DIFF
--- a/src/connection.rs
+++ b/src/connection.rs
@@ -38,6 +38,7 @@ use crate::arena;
 use crate::buffer::{
     Buffer, ContiguousBuffer, LimitBufsMut, TmpBuffer, VecRingBuffer, VECTORED_MAX,
 };
+use crate::counter::Counter;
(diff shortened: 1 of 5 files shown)


src/connection.rs
madeline-boby, line 1: blocks_avail may not need to be enclosed in an Option, but it should be okay for now

ghost (approved):

PR #47754, opened 2023-10-11
drop message before receiving, and don't obscure tracking error

jkarneges:
This fixes a bug where if a WebSocket request is rejected, we might attempt to receive from a channel while holding onto a previous message, resulting in an error and disconnecting the client without sending the rejection response.

Also, the error is incorrectly reported as BadMessage, when it should be ValueActive. Returning the correct error here is important as ValueActive is considered a program logic error and is logged at a higher log level than other errors.


sima-fastly (approved):
this looks good to me. has the fix been tested on a cache node?

jkarneges:
It will be. For now tested locally.

PR #47755, opened 2023-10-14
runner logging using simple logger

sima-fastly:
This is the 9th PR in a series of PRs to rewrite Pushpin's Runner in Rust.
here we are :
-setting simple logger as runners logger
-logging each service logs on their own log level, target ... 
-spawning threads to capture stdout and stderr per service and pipe to simple logger and finally joining those threads
-setting up a simple logger so it doesn't double log the level and timestamps. 

This was tested manually 
Screenshot 2023-10-16 at 2 54 02 PM

### DIFF
--- a/src/service.rs
+++ b/src/service.rs
@@ -383,22 +383,11 @@ fn start_log_handler(
 }
 
 fn log_message(name: &str, level: log::Level, msg: &str) {
-    let mut msg_to_log = msg.to_string();
-
-    let pieces: Vec<&str> = msg.splitn(4, ' ').collect();
-    if pieces.len() == 4 && pieces[0].starts_with('[') {
-        msg_to_log = pieces[3].to_string();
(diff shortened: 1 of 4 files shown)


src/service.rs
```diff
@@ -288,22 +319,75 @@ impl PushpinHandlerService {
...
+    result.push(Some(thread::spawn(move || {
+        let reader_err = BufReader::new(stderr);
+        for line in reader_err.lines() {
+            match line {
+                Ok(msg) => log_message(&name, log::Level::Error, &msg),
+                Err(_) => {
+                    log_message(
+                        &name,
+                        log::Level::Error,
+                        "failed to read from standard out.",
```
jkarneges, line 374: "standard error"

jkarneges (approved):

src/service.rs
sima-fastly: thanks for catching this, updating now

PR #47756, opened 2023-10-19
Runner file logging

sima-fastly:
This is the 10th PR in a series of PRs to rewrite Pushpin's Runner in Rust.
here we are :
-Simplifying runner logger logic 
-Adding the option to log to file if --logfile cli param is set
-Ensuring log dir & file is created and available for writing

This was tested manually
Screenshot 2023-10-19 at 5 32 03 AM

### DIFF
--- a/src/bin/pushpin.rs
+++ b/src/bin/pushpin.rs
@@ -16,7 +16,7 @@
 
 use clap::Parser;
 use log::{info, LevelFilter};
-use pushpin::runner::{get_runner_logger, ArgsData, CliArgs, Settings};
+use pushpin::runner::{get_log_file, get_runner_logger, ArgsData, CliArgs, Settings};
 use pushpin::service::start_services;
 use std::error::Error;
 use std::process;
@@ -25,7 +25,7 @@ fn process_args_and_run(args: CliArgs) -> Result<(), Box<dyn Error>> {
     let args_data = ArgsData::new(args)?;
     let settings = Settings::new(args_data)?;
 
-    log::set_logger(get_runner_logger(settings.log_file.clone())).unwrap();
+    log::set_logger(get_runner_logger(get_log_file(settings.log_file.clone()))).unwrap();
     let ll = settings
         .log_levels
         .get("")
--- a/src/bin/pushpin.rs
+++ b/src/bin/pushpin.rs
@@ -15,7 +15,7 @@
  */
 
 use clap::Parser;
-use log::{info, LevelFilter};
+use log::{error, info, LevelFilter};
 use pushpin::runner::{get_runner_logger, open_log_file, ArgsData, CliArgs, Settings};
 use pushpin::service::start_services;
 use std::error::Error;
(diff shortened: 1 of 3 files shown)


src/runner.rs
```diff
@@ -909,47 +928,12 @@ impl Log for RunnerLogger {
...
-            println!(
-                "{} {} {} [{}] {}",
-                msg_parts[0],
-                msg_parts[1],
-                msg_parts[2],
-                record.target(),
-                msg_parts[3..].join(" ")
-            );
+        if let Some(log_file) = &self.log_file {
+            if let Ok(mut file) = log_file.write() {
```
jkarneges, line 932: Acquiring the lock will only fail due to lock poisoning which we can treat as a logic error, so we can unwrap/expect here instead of a conditional.

src/runner.rs
```diff
@@ -958,15 +942,12 @@ impl Log for RunnerLogger {
...
+pub fn get_runner_logger(log_file_path: Option<PathBuf>) -> &'static impl Log {
     static INIT: Once = Once::new();
 
     unsafe {
         INIT.call_once(|| {
-            let local_offset =
-                UtcOffset::current_local_offset().expect("failed to get local time offset");
-
-            LOGGER.write(RunnerLogger { local_offset });
+            LOGGER.write(RunnerLogger::new(log_file_path).expect("failed to set logger"));
```
jkarneges, line 950: Let's not panic if the runner can't be initialized, and instead handle the Result.

src/service.rs
```diff
@@ -383,11 +383,24 @@ fn start_log_handler(
...
+        .map(|(i, _)| i)
+        .unwrap_or_else(|| 0);
+
     log::logger().log(
         &log::Record::builder()
             .level(level)
             .target(name)
-            .args(format_args!("{}", msg))
+            .args(format_args!(
+                "{} [{}]{}",
```
jkarneges, line 399: Might need another space in there?

src/runner.rs
```diff
@@ -909,66 +907,44 @@ impl Log for RunnerLogger {
...
-            println!(
-                "{} {} {} [{}] {}",
-                msg_parts[0],
-                msg_parts[1],
-                msg_parts[2],
-                record.target(),
-                msg_parts[3..].join(" ")
-            );
+        if let Some(log_file) = &self.log_file {
+            let mut log_file: MutexGuard<'_, File> = log_file.lock().unwrap();
```
jkarneges, line 911: nit: you can get rid of the explicit `MutexGuard` type here and let type inference figure it out.

src/service.rs
sima-fastly: no space needed here as the msg includes a space

src/bin/pushpin.rs
```diff
@@ -25,7 +25,17 @@ fn process_args_and_run(args: CliArgs) -> Result<(), Box<dyn Error>> {
     let args_data = ArgsData::new(args)?;
     let settings = Settings::new(args_data)?;
 
-    log::set_logger(get_runner_logger()).unwrap();
+    let logger = match settings.log_file.clone() {
+        Some(x) => {
+            let log_file = match open_log_file(x) {
+                Ok(x) => Some(x),
+                Err(_) => None,
```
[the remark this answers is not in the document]
sima-fastly: added a log line to indicate the err to open log file and then default to logging to std out.

src/service.rs
jkarneges: ah yes

jkarneges (approved):

PR #47759, opened 2023-10-27
import qzmq files and remove submodule

jkarneges:
Currently, we use zmq from both Rust and C++ code, but each language uses a different process for detecting and linking against zmq (C++ uses the configure script, and Rust uses the logic from the zmq crate), which is a) weird and could lead to inconsistent results, and b) problematic when mixing Rust & C++ code together. In order to fix this, I'm planning to have the C++ code access zmq through Rust rather than directly. This way, only the Rust code will contain direct zmq usage and we can keep the detecting/linking logic in one place (in Rust). Relatedly, this will require turning all of our C++ programs into Rust programs, even if the Rust versions are just tiny wrappers around C++, to get Rust's linking behavior in all of our programs.

For our C++ zmq code to go through Rust, we'll need to be able to modify the qzmq dependency which we currently pull in as a submodule. So, as a first step, this PR copies qzmq into in the repo so we can modify it. The rest of the above plan will be handled in subsequent PRs.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,0 @@
-[submodule "qzmq"]
-	path = src/corelib/qzmq
-	url = https://github.com/jkarneges/qzmq.git
--- a/src/corelib/qzmq/.gitignore
+++ b/src/corelib/qzmq/.gitignore
@@ -1,0 +1,7 @@
+conf.pri
+Makefile
+*.o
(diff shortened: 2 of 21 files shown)


sima-fastly (approved):
all looks good assuming it has been tested on a cache node

PR #47760, opened 2023-10-27
update zmq crate

jkarneges:


sima-fastly (approved):
LGTM if all queue functionalities are tested out.

PR #47762, opened 2023-10-30
qt 6 compatibility

jkarneges:


configure
```diff
@@ -737,10 +737,15 @@ QStringList qc_pathlist()
     QStringList list;
     QString     path = qc_getenv("PATH");
     if (!path.isEmpty()) {
+#if QT_VERSION >= 0x060000
```
sima-fastly, line 740: where does that qt version is coming from? 
0x060000 in decimal is `393216`

jkarneges: it's intentionally hex: https://doc.qt.io/qt-5/qtglobal.html#QT_VERSION

sima-fastly (approved):
lgtm

PR #47764, opened 2023-11-01
fix qt 5.12 & 5.13 compatibility

jkarneges:
The qt 6 compatibility changes accidentally broke older 5.x compat, notably 5.12 which is used by Ubuntu 20.04.


sima-fastly (approved):
lgtm 🎖️

PR #47765, opened 2023-11-01
pushpin binary should be in the root dir

jkarneges:


sima-fastly (approved):
thanks for this update

PR #47766, opened 2023-11-02
wrap zmq crate with libzmq-like C API and use from qzmq

jkarneges:
This change makes the C++ and Rust code share the same libzmq (the one used by the zmq crate), to avoid depending on two different versions of libzmq at the same time.

To simplify the way we handle dependencies, the C++ code does not interface directly with the libzmq used by the zmq crate. Instead, we wrap the zmq crate with a libzmq-like C API that the C++ code can use.

Testing: the tests pass (there's a lot of zmq in there) and also tested locally.


sima-fastly (approved):
LGTM

PR #47767, opened 2023-11-03
rename src/corelib to src/cpp

jkarneges:
this is preparation for consolidating all C++ code into src/cpp


sima-fastly (commented):
there are some refs to corelib in `.gitignore` and `package.sh`

jkarneges:
Good catch. Fixed.

sima-fastly (approved):
lgtm

PR #47768, opened 2023-11-03
move proxy code into src/cpp

jkarneges:
This change moves the proxy code and its tests into the cpp lib, renames its `main` function to `proxy_main`, and reduces src/proxy to a small C program that simply calls `proxy_main` from the lib.

After this lands, the plan is to make similar changes for all of our other C++ programs such that all of our C++ code ends up consolidated in src/cpp wrapped by small C mains. Then after that we can replace the C mains with Rust mains, such that all of our programs become Rust programs at the top level.


sima-fastly (approved):
great initiative 
if all functionalities are tested, the code looks good to me.

jkarneges:
Yep, all tests pass and also manually tested.

PR #47771, opened 2023-11-07
convert all programs to rust programs calling mains from cpp lib

jkarneges:
Notably this changes the build order. Previously, the Rust code was built first (lib and programs) followed by all the C++ code which could use the Rust lib as needed. With this change, the C++ lib is built first, followed by the Rust code which uses the C++ lib as needed, followed by the C++ tests which use both the C++ lib and Rust lib as needed.


src/rust/rust.pro
```diff
@@ -34,9 +50,25 @@ condure_bin.target = $$bin_dir/condure
 condure_bin.depends = condure_build
 condure_bin.commands = mkdir -p $$bin_dir && cp -a $$target_dir/condure $$bin_dir/condure
 
-pushpin_bin.target = $$root_dir/pushpin
-pushpin_bin.depends = pushpin_build
-pushpin_bin.commands = cp -a $$target_dir/pushpin $$root_dir/pushpin
+m2adapter_bin.target = $$bin_dir/m2adapter
+m2adapter_bin.depends = m2adapter_build
+m2adapter_bin.commands = mkdir -p $$bin_dir && cp -a $$target_dir/m2adapter $$bin_dir/m2adapter
```
sima-fastly, line 55: do we need `mkdir -p $$bin_dir` cmd to be repeated for these bins?

jkarneges: Sort of. The bin dir isn't part of the repo and is created on demand. Since it is possible for make to run individual targets or to run these targets in different orders, they all try to make the bin dir.

sima-fastly: in that case do we need to add the `mkdir` to runner_legacy_bin.commands & runner_bin.commands ?

sima-fastly (approved):

src/rust/rust.pro
jkarneges: ah, the runner(s) go in the root dir not the bin dir, so they don't need it.

PR #47773, opened 2023-11-07
remove check for libzmq in configure script

jkarneges:
the check is not needed since the zmq crate (as of 0.10) vendors libzmq and all zmq usage goes through the crate


sima-fastly (approved):
lgtm

PR #47775, opened 2023-11-09
build the c++ tests into a library and run them from rust

jkarneges:
This lets us run the C++ tests with `cargo test` and also avoids needing to build the C++ test programs after the Rust lib is built.

The C++ tests add some noise to the test output, since Rust's test output capture mechanism only works on Rust I/O calls. We can look at cleaning this up later.


src/cpp/rtimer.cpp
```diff
@@ -244,4 +246,10 @@ void RTimer::init(int capacity)
 	g_manager = new TimerManager(capacity);
 }
 
+void RTimer::deinit()
+{
+	delete g_manager;
+	g_manager = 0;
```
maddie-boby, line 252: Is there a reason we use 0 instead of NULL or nullptr?

jkarneges: For consistency with the rest of the code, which uses 0 for null due to originally using a pre-C++11 dialect.

Would not be opposed to a PR that changes all pointer nullings to nullptr. :)

madeline-pratt (approved):

PR #47779, opened 2023-11-10
Replacing Qt Signal Slot with Boost Signals2

sima-fastly:
This is the first PR in a series of PRs to replace Qt signal & slot system with Boost Signals2.

In this PR we are replacing qt signal slots in the very top layer of all services quit & hup signals.

Tested manually 
Screenshot 2023-11-14 at 5 32 26 AM
Screenshot 2023-11-14 at 5 34 34 AM
Screenshot 2023-11-14 at 5 34 42 AM


src/cpp/handler/handlermain.cpp
```diff
@@ -52,8 +51,8 @@ int handler_main(int argc, char **argv)
 {
 	QCoreApplication qapp(argc, argv);
 
-	HandlerAppMain appMain;
-	QTimer::singleShot(0, &appMain, SLOT(start()));
+	HandlerAppMain appMain;	
+	QTimer::singleShot(0, [&appMain]() {appMain.start();});
```
jkarneges, line 55: Nice find.

src/cpp/proxy/main.cpp
```diff
@@ -53,8 +52,8 @@ int proxy_main(int argc, char **argv)
 	QCoreApplication qapp(argc, argv);
 
 	AppMain appMain;
-	QTimer::singleShot(0, &appMain, SLOT(start()));
-	return qapp.exec();
+	QTimer::singleShot(0, [&appMain]() {appMain.start();});
+	return qapp.exec();
```
jkarneges, line 56: Whitespace nit.

jkarneges (commented):
We should also remove the `Q_OBJECT` macro from class definitions as well as the `QObject` inheritance.

qcm/boost.qcm
```diff
@@ -0,0 +1,61 @@
...
+section: libraries
+description: Configuration for the Boost.Signals2 library, a part of the Boost C++ Libraries.
+-----END QCMOD-----
+*/
+
+class qc_boost : public ConfObj
+{
+public:
+    qc_boost(Conf *c) : ConfObj(c) {}
+    QString name() const { return "Boost.Signals2 >= 1.83"; }
```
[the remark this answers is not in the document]
sima-fastly: 1.71 was throwing some weird type conversion errors with Qt objects, that's why I picked the latest.

src/cpp/handler/handlerapp.cpp
```diff
@@ -152,14 +152,16 @@ class HandlerApp::Private : public QObject
 	HandlerApp *q;
 	ArgsData args;
 	HandlerEngine *engine;
+	boost::signals2::connection quitConnection;
+	boost::signals2::connection hupConnection;
```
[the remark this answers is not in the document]
sima-fastly: yes, to gracefully close connections when the service is stopping

src/cpp/handler/handlerapp.h
```diff
@@ -35,8 +35,7 @@ class HandlerApp : public QObject
 
 	void start();
 
-signals:
-	void quit(int returnCode = 0);
+	boost::signals2::signal<void(int)> quit;
```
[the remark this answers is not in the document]
sima-fastly: I can do sth like `using connection = boost::signals2::connecection;`

src/cpp/handler/handlermain.cpp
sima-fastly: the `make` process found it for me

sima-fastly:
> We should also remove the `Q_OBJECT` macro from class definitions as well as the `QObject` inheritance.

I agree, that can be done after Qt signal/slot cleanup?

jkarneges:
> > We should also remove the `Q_OBJECT` macro from class definitions as well as the `QObject` inheritance.
> 
> I agree, that can be done after Qt signal/slot cleanup?

That's fine but do we know if such removal is possible? It only makes sense to move to boost if we can actually get off of QObject. It would be good to try this locally at least, and then the change could go in a later PR.

qcm/boost.qcm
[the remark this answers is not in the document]
jkarneges: Hmm, let's investigate this some more. Maybe there's a different way to write the code that would work with the older version.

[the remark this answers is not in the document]
sima-fastly: done

sima-fastly:
> > > We should also remove the `Q_OBJECT` macro from class definitions as well as the `QObject` inheritance.
> > 
> > 
> > I agree, that can be done after Qt signal/slot cleanup?
> 
> That's fine but do we know if such removal is possible? It only makes sense to move to boost if we can actually get off of QObject. It would be good to try this locally at least, and then the change could go in a later PR.

from what I understand so far, I am fairly certain it is possible but I am gonna have to peel back layers one by one which will take weeks.

sima-fastly:
> > > > We should also remove the `Q_OBJECT` macro from class definitions as well as the `QObject` inheritance.
> > > 
> > > 
> > > I agree, that can be done after Qt signal/slot cleanup?
> > 
> > 
> > That's fine but do we know if such removal is possible? It only makes sense to move to boost if we can actually get off of QObject. It would be good to try this locally at least, and then the change could go in a later PR.
> 
> from what I understand so far, I am fairly certain it is possible but I am gonna have to peel back layers one by one which will take weeks.

thanks for this suggestion, QObject & Q_Object are removed from services' main fn

src/cpp/handler/handlermain.cpp
```diff
@@ -24,18 +24,15 @@
...
 {
-	Q_OBJECT
-
 public:
 	HandlerApp *app;
 
-public slots:
 	void start()
 	{
-		app = new HandlerApp(this);
```
jkarneges, line 37: one drawback to not using QObject is we don't get automatic cleanup. before, this parenting would automatically destroy the `HandlerApp` when the `HandlerAppMain` itself was destroyed. probably we'll want to move to some other non-Qt solution for this, but in the meantime lets just clean things up manually with a destructor. we should also disconnect the signal there too, prior to destruction, e.g.:

```c++
~HandlerAppMain()
{
        quitConnection.disconnect();
        delete app;
}
```

sima-fastly: `quicConnection.disconnect()` & `hupConnection.disconnect()` can be added to `~HandlerApp()` 
and then maybe call `~HandlerApp()` from within `~HandlerAppMain()` ?

src/cpp/handler/handlermain.cpp
```diff
@@ -44,6 +41,12 @@ public slots:
 		delete app;
 		QCoreApplication::exit(returnCode);
 	}
+	
+	~HandlerAppMain()
+	{
+		app->~HandlerApp();
```
jkarneges, line 47: C++ destructors shouldn't be called directly. The `delete app` below will do it.

jkarneges (approved):

PR #47782, opened 2023-11-14
invoke qmake+make from build.rs to build c++ code

jkarneges:
Further cargo-ification of the build. Instead of `make` recursively running `make` in src/cpp to build the C++ code prior to `cargo build`, `make` simply runs `cargo build` which handles everything (`build.rs` runs `make` in src/cpp). This means `cargo build` will build all the code, both C++ and Rust.

Notably, this decouples the inner and outer make processes, and also removes cargo's dependency on the outer make process. All cargo depends on now is the `conf.pri` file generated by the configure script.

From here, we could consider:

* using the `cc` crate to build the C++ code instead of make
* replacing the configure script with something not qt-based, or moving the configuration logic into `build.rs` and not having a configure script at all

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -4,6 +4,7 @@ use std::error::Error;
 use std::fs;
 use std::io::{BufRead, BufReader};
 use std::path::{Path, PathBuf};
+use std::process::Command;
 
 fn main() -> Result<(), Box<dyn Error>> {
     let conf = {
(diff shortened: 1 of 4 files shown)


build.rs
```diff
@@ -53,14 +77,6 @@ fn main() -> Result<(), Box<dyn Error>> {
 
     println!("cargo:rerun-if-changed=conf.pri");
     println!("cargo:rerun-if-changed=src");
-    println!(
```
sima-fastly, line 56: why are we removing these?

jkarneges: Ah, their presence causes `cargo build` to want to rebuild again, because those libs are now created after cargo starts running rather than before. In any case they shouldn't be needed since any change in src is sufficient to ensure a rebuild.

sima-fastly (approved):

PR #47786, opened 2023-11-16
install bins from pushpin.pro and remove unused lib.pri

jkarneges:
We would like to manage the project with cargo as much as possible, however unlike qmake it does not support post-build logic. This PR starts to tease apart the build logic and post-build logic, with the eventual goal of letting the build logic be entirely managed by cargo. The post-build logic will continue to be managed outside of cargo.

Tested the install locally.


src/pushpin/pushpin.pro
```diff
@@ -2,6 +2,9 @@ TEMPLATE = aux
 
 include($$OUT_PWD/../../conf.pri)
 
+bin_dir = $$PWD/../../bin
+root_dir = $$PWD/../..
```
sima-fastly, line 6: this is a bit neater : 
`bin_dir =  $$root_dir/bin`

sima-fastly (approved):

PR #47801, opened 2023-11-27
generate individual c++ makefiles in target dir

jkarneges:
This generates the c++ makefiles in the target dir rather than under src/cpp, so no files are written under src/cpp during build, and so `cargo clean` will remove all intermediate build files including generated makefiles. As a further simplification, the makefiles are generated individually via multiple qmake invocations rather than by relying on a qmake subdirs project. Also fix make install and move the location of postbuild's pri file to avoid warnings when running make clean.


.gitignore
```diff
@@ -21,11 +21,9 @@ vendor
 /conf.log
 /conf.pri
 /Makefile
-/src/cpp/Makefile
-/src/cpp/cpp/Makefile
-/src/cpp/tests/Makefile
 /src/runner/certs/*.crt
 /src/runner/certs/*.key
+/postbuild/conf.pri
```
sima-fastly, line 26: if we want all conf.pri files to be ignore we can add `conf.pri` once in .gitignore

sima-fastly (approved):

PR #47803, opened 2023-11-28
remove configure script and base qmake project

jkarneges:
This removes the configure script and qconf modules, and replaces the base qmake project with a hand-written makefile. Support for conf.pri is removed and any configuration must be done via environment variables. The project can now be built out of the box simply by running `make` (no need to run `qmake` first).


sima-fastly (approved):
this is so neat! thanks 🎖️

sima-fastly (commented):
do we need to add these makefiles to .gitignore? 
Screenshot 2023-11-27 at 8 33 53 PM

jkarneges:
Ah, those makefiles are old, you can delete them. They are now generated in the target dir which is ignored.

PR #47807, opened 2023-11-28
check boost version and add to include path

sima-fastly:
check boost version and add to include path in `build.rs`

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -73,10 +73,14 @@ fn env_or_default(name: &str, defaults: &HashMap<String, String>) -> String {
     }
 }
 
-fn write_cpp_conf_pri(path: &Path, _boost_include_path: &str) -> Result<(), Box<dyn Error>> {
+fn write_cpp_conf_pri(path: &Path, boost_include_path: &str) -> Result<(), Box<dyn Error>> {
     let mut f = fs::File::create(path)?;
 
     writeln!(&mut f)?;
+    println!(
+        "======= boost_include_path : {} =======",
+        boost_include_path
+    );
     writeln!(&mut f, "INCLUDEPATH += /usr/local/include")?;
 
     Ok(())
@@ -202,11 +206,10 @@ fn main() -> Result<(), Box<dyn Error>> {
         fs::create_dir_all(cpp_out_dir.join(dir))?;
     }
 
-    let mut _boost_include_path = String::new();
+    let mut boost_include_path = String::new();
     match check_boost_version() {
         Ok(true) => {
-            _boost_include_path =
-                find_boost_include_path().ok_or("Boost include path not found")?;
+            boost_include_path = find_boost_include_path().ok_or("Boost include path not found")?;
         }
         Ok(false) => {
             return Err("Boost version is not sufficient.".to_string().into());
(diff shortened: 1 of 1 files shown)


build.rs
```diff
@@ -101,6 +103,38 @@ fn write_postbuild_conf_pri(
...
+
+    let major_version = version_parts[0].parse::<i32>()?;
+    let minor_version = version_parts[1].parse::<i32>()?;
+
+    Ok(major_version > 1 || (major_version == 1 && minor_version >= 71))
+}
+
+fn find_boost_include_path() -> Option<String> {
+    let possible_paths = vec!["/usr/local/include", "/usr/include"];
+    let boost_header = "boost/signals2.hpp";
```
jkarneges, line 126: How about searching for version.hpp here instead of signals2.hpp, and then reading the version number out of the found file, rather than doing the version check separately in `check_boost_version` ? That would help ensure the path being version checked is the same path being used.

Also try to use Rust code to parse the file rather than shell commands.

BTW, for checking the version string, you can reuse the `check_version` function.

sima-fastly: great points! 
I have updated the PR to include one `get_boost_path` fn where boost version is read from version.hpp file and verified by `check_verison` func.

build.rs
```diff
@@ -101,6 +102,43 @@ fn write_postbuild_conf_pri(
...
+            for line in reader.lines() {
+                match line {
+                    Ok(x) => {
+                        if x.contains("#define BOOST_LIB_VERSION") {
+                            let parts: Vec<&str> = x.split('"').collect();
+                            if parts.len() >= 2 {
+                                let version = parts[1].replace('_', ".");
+                                check_version("boost", &version, 1, 71)?;
+                            } else {
+                                return Err("Error finding boost package verion1"
```
jkarneges, line 124: typo

build.rs
```diff
@@ -101,6 +102,43 @@ fn write_postbuild_conf_pri(
...
+                                    .to_string()
```
jkarneges, line 125: I think you can leave out to_string() on these errors and into() will take care of it

build.rs
```diff
@@ -77,7 +77,8 @@ fn write_cpp_conf_pri(path: &Path) -> Result<(), Box<dyn Error>> {
     let mut f = fs::File::create(path)?;
 
     writeln!(&mut f)?;
-    writeln!(&mut f, "INCLUDEPATH = /usr/local/include")?;
+    #[cfg(target_os = "macos")]
+    writeln!(&mut f, "INCLUDEPATH += {}", get_boost_path()?)?;
```
jkarneges, line 81: Instead of making this mac-only, I think what you want is to only add the path if it isn't /usr/include?

build.rs
```diff
@@ -77,7 +77,11 @@ fn write_cpp_conf_pri(path: &Path) -> Result<(), Box<dyn Error>> {
     let mut f = fs::File::create(path)?;
 
     writeln!(&mut f)?;
-    writeln!(&mut f, "INCLUDEPATH = /usr/local/include")?;
+    if let Ok(boost_path) = get_boost_path() {
```
jkarneges, line 80: This looks like it will ignore errors?

sima-fastly: ooh sorry! wanted to write a fancy clause, not realizing the full effect. 
fix is in

jkarneges (approved):

PR #47817, opened 2023-12-04
boostification of simple http server

sima-fastly:
simplehttpserver's signals updated from qt signals to boost. 
all the corresponding connections were also updated.

Tested locally

### DIFF
--- a/src/cpp/simplehttpserver.cpp
+++ b/src/cpp/simplehttpserver.cpp
@@ -495,7 +495,7 @@ private:
 			QLocalSocket *sock = ((QLocalServer *)server)->nextPendingConnection();
 			SimpleHttpRequest *req = new SimpleHttpRequest(maxHeadersSize, maxBodySize);
 			connect(req->d, &SimpleHttpRequest::Private::ready, this, &SimpleHttpServerPrivate::req_ready);
-			finishedConnection = req->finished.connect(boost::bind(&SimpleHttpServerPrivate::req_finished, this));
+			finishedConnection = req->finished.connect(boost::bind(&SimpleHttpServerPrivate::req_finished, this, req));
 			accepting += req;
 			req->d->start(sock);
 		}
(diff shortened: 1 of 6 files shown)


src/cpp/statsmanager.cpp
```diff
@@ -1556,8 +1556,8 @@ private slots:
 			).arg(prometheusPrefix, m.name, m.help, prometheusPrefix, m.name, m.type, prometheusPrefix, m.name, value.toString());
 		}
 
-		connect(req, &SimpleHttpRequest::finished, req, &QObject::deleteLater);
-
+		req->finished.connect(boost::bind(&SimpleHttpRequest::deleteLater, req));
+
```
jkarneges, line 1560: whitespace

src/cpp/simplehttpserver.cpp
```diff
@@ -486,15 +487,15 @@ class SimpleHttpServerPrivate : public QObject
...
+private:
 	void server_newConnection()
 	{
 		if(local)
 		{
 			QLocalSocket *sock = ((QLocalServer *)server)->nextPendingConnection();
 			SimpleHttpRequest *req = new SimpleHttpRequest(maxHeadersSize, maxBodySize);
 			connect(req->d, &SimpleHttpRequest::Private::ready, this, &SimpleHttpServerPrivate::req_ready);
-			connect(req, &SimpleHttpRequest::finished, this, &SimpleHttpServerPrivate::req_finished);
+			finishedConnection = req->finished.connect(boost::bind(&SimpleHttpServerPrivate::req_finished, this, req));
```
jkarneges, line 498: We need to keep track of a connection per request here, or else every new request overwrites the one connection.

src/cpp/simplehttpserver.cpp
```diff
@@ -553,7 +553,11 @@ SimpleHttpRequest *SimpleHttpServer::takeNext()
 	if(!d->pending.isEmpty())
 	{
 		SimpleHttpRequest *req = d->pending.takeFirst();
-		req->disconnect(d);
+		auto conn = d->finishedConnections.find(req);
+		if (conn != d->finishedConnections.end()) {
+			conn->second.disconnect();
```
jkarneges, line 558: Since these are scoped connections, maybe deleting the entry is enough and this call is not needed?

jkarneges (approved):

src/cpp/simplehttpserver.cpp
sima-fastly: very true! done

src/cpp/simplehttpserver.cpp
```diff
@@ -553,7 +553,10 @@ SimpleHttpRequest *SimpleHttpServer::takeNext()
 	if(!d->pending.isEmpty())
 	{
 		SimpleHttpRequest *req = d->pending.takeFirst();
-		req->disconnect(d);
+		auto conn = d->finishedConnections.find(req);
+		if (conn != d->finishedConnections.end())
+			d->finishedConnections.erase(conn);
```
jkarneges, line 558: You may even be able to write this more simply as: `d->finishedConnections.erase(req)`.

jkarneges (approved):

PR #47820, opened 2023-12-06
proxy req session zhttpReq readyread disconnect fix

sima-fastly:
proxy req session zhttpReq readyread disconnect fix
the disconnect needs to happen per request 

tested locally


sima-fastly:
> Hmm, I don't think this code needs to track multiple connections since there is only one inner ZhttpRequest per RequestSession.

that's a great point! no need for this update then. 
ill be closing this PR

PR #47826, opened 2023-12-08
enable unsound time zone checking, and fall back to utc

jkarneges:
M1/docker fix, adapted from https://github.com/fanout/condure/pull/15


sima-fastly (approved):

sima-fastly (commented):
was this tested on an M1?

jkarneges:
Not yet. I think it's the right thing to be doing regardless, so I've merged, but we'll want to confirm.

PR #47827, opened 2023-12-08
boostification of zrpcrequest

sima-fastly:
boostification of zrpcrequest

tested locally

### DIFF
--- a/src/cpp/proxy/requestsession.cpp
+++ b/src/cpp/proxy/requestsession.cpp
@@ -191,7 +191,8 @@ public:
 	Connection pausedConnection;
 	Connection readyReadConnection;
 	Connection bytesWrittenConnection;
-	Connection finishedConnection;
+	Connection inspectFinishedConnection;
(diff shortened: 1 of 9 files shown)


src/cpp/proxy/requestsession.cpp
```diff
@@ -852,7 +853,7 @@ class RequestSession::Private : public QObject
 			adata.channelPrefix = route.prefix;
 
 			acceptRequest = new AcceptRequest(acceptManager, this);
-			connect(acceptRequest, &AcceptRequest::finished, this, &Private::acceptRequest_finished);
+			finishedConnection = acceptRequest->finished.connect(boost::bind(&Private::acceptRequest_finished, this));
```
jkarneges, line 856: This is a different signal/slot pair than the one for inspectRequest. I think the two connections don't occur at the same time so it could be fine to reuse `finishedConnection` for both, but it having different connection members for each might be less error prone.

jkarneges (approved):

PR #47830, opened 2023-12-09
build the c++ code in the correct mode

jkarneges:
Notably this places the build files within mode-specific dirs (e.g. `target/debug`).


sima-fastly (approved):
lgtm

PR #47831, opened 2023-12-11
boostificaiton of rtimer

sima-fastly:
boostificaiton of rtimer

tested locally

### DIFF
--- a/src/cpp/handler/httpsession.cpp
+++ b/src/cpp/handler/httpsession.cpp
@@ -182,6 +182,8 @@ public:
 	Connection pausedConnection;
 	Connection readyReadOutConnection;
 	Connection errorOutConnection;
+	Connection timeoutConnection;
+	Connection retryTimerConneciton;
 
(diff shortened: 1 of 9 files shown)


src/cpp/handler/httpsession.cpp
```diff
@@ -210,10 +212,10 @@ class HttpSession::Private : public QObject
 		errorConnection = req->error.connect(boost::bind(&Private::req_error, this));
 
 		timer = new RTimer(this);
-		connect(timer, &RTimer::timeout, this, &Private::timer_timeout);
+		timeoutConnection = timer->timeout.connect(boost::bind(&Private::timer_timeout, this));
 
 		retryTimer = new RTimer(this);
-		connect(retryTimer, &RTimer::timeout, this, &Private::retryTimer_timeout);
+		retryTimerConneciton = retryTimer->timeout.connect(boost::bind(&Private::retryTimer_timeout, this));
 		retryTimer->setSingleShot(true);
```
jkarneges, line 219: The original code has explicit disconnects for these too so we'll need to retain that behavior.

sima-fastly: updated to boost disconnect

jkarneges (approved):

PR #47832, opened 2023-12-12
boostification of domainmap cpp

sima-fastly:
boostification of domainmap cpp

tested locally

### DIFF
--- a/src/cpp/proxy/domainmap.cpp
+++ b/src/cpp/proxy/domainmap.cpp
@@ -305,6 +305,11 @@ public:
 	Signal changed;
 
 public slots:
+	void doChanged()
+	{
+		changed();
+	}
+
 	void start()
 	{
 		if(!fileName.isEmpty())
@@ -708,8 +713,6 @@ class DomainMap::Thread : public QThread
 {
 	Q_OBJECT
 
-	Connection startedConnection;
-
 public:
 	QString fileName;
 	Worker *worker;
@@ -733,7 +736,7 @@ public:
 	{
 		worker = new Worker;
 		worker->fileName = fileName;
-		startedConnection = worker->started.connect(boost::bind(&Thread::worker_started, this));
+		Connection startedConnection = worker->started.connect(boost::bind(&Thread::worker_started, this));
 		QMetaObject::invokeMethod(worker, "start", Qt::QueuedConnection);
 		exec();
 		delete worker;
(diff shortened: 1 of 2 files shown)


src/cpp/proxy/domainmap.cpp
```diff
@@ -732,13 +733,13 @@ class DomainMap::Thread : public QThread
 	{
 		worker = new Worker;
 		worker->fileName = fileName;
-		connect(worker, &Worker::started, this, &Thread::worker_started, Qt::DirectConnection);
+		startedConnections[worker] = worker->started.connect(boost::bind(&Thread::worker_started, this));
 		QMetaObject::invokeMethod(worker, "start", Qt::QueuedConnection);
 		exec();
 		delete worker;
```
jkarneges, line 739: Keeping a map shouldn't be needed here, since the `run()` method runs synchronously from start to finish (it is invoked by `QThread`), and the worker is destroyed at the end.

src/cpp/proxy/domainmap.cpp
```diff
@@ -773,13 +776,13 @@ class DomainMap::Private : public QObject
 		thread->start();
 
 		// worker guaranteed to exist after starting
-		connect(thread->worker, &Worker::changed, this, &Private::doChanged);
+		changedConnections[thread] = thread->worker->changed.connect(boost::bind(&Private::doChanged, this));
```
jkarneges, line 779: This map shouldn't be needed, since the `start()` method is only called once.

src/cpp/proxy/domainmap.cpp
```diff
@@ -285,7 +285,7 @@ class DomainMap::Worker : public QObject
 
 		log_info("routes loaded with %d entries", allRules.count());
 
-		QMetaObject::invokeMethod(this, "changed", Qt::QueuedConnection);
+		QMetaObject::invokeMethod(this, "doChanged", Qt::QueuedConnection);
```
jkarneges, line 288: DomainMap::Worker does not have a `doChanged` slot.

src/cpp/proxy/domainmap.cpp
```diff
@@ -709,6 +708,8 @@ class DomainMap::Thread : public QThread
 {
 	Q_OBJECT
 
+	Connection startedConnection;
```
jkarneges, line 711: The connection only needs to exist within `run()` so it doesn't need to be a class member.

src/cpp/proxy/domainmap.cpp
sima-fastly: thanks for catching this one!

jkarneges (approved):

PR #47838, opened 2023-12-13
boostification of ws over http

sima-fastly:
boostification of ws over http

tested locally

### DIFF
--- a/src/cpp/proxy/websocketoverhttp.cpp
+++ b/src/cpp/proxy/websocketoverhttp.cpp
@@ -95,6 +95,7 @@ public:
 private:
 	void cleanupSocket(WebSocketOverHttp *sock)
 	{
+		disconnectedConnection.erase(sock);
 		delete sock;
 	}
 
--- a/src/cpp/proxy/websocketoverhttp.cpp
+++ b/src/cpp/proxy/websocketoverhttp.cpp
@@ -69,6 +69,8 @@ class WebSocketOverHttp::DisconnectManager : public QObject
 {
 	Q_OBJECT
 
+	map<WebSocketOverHttp *, Connection> disconnectedConnection;
+
 public:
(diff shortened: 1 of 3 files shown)


src/cpp/proxy/websocketoverhttp.cpp
```diff
@@ -78,7 +80,7 @@ class WebSocketOverHttp::DisconnectManager : public QObject
 	void addSocket(WebSocketOverHttp *sock)
 	{
 		sock->setParent(this);
-		connect(sock, &WebSocketOverHttp::disconnected, this, &DisconnectManager::sock_disconnected);
+		disconnectedConnection[sock] = sock->disconnected.connect(boost::bind(&DisconnectManager::sock_disconnected, this, sock));
```
jkarneges, line 83: The connection needs to be removed from the map too, probably in `cleanupSocket`.

src/cpp/proxy/wsproxysession.cpp
```diff
@@ -530,7 +531,7 @@ class WsProxySession::Private : public QObject
 				if(target.oneEvent)
 					woh->setMaxEventsPerRequest(1);
 
-				connect(woh, &WebSocketOverHttp::aboutToSendRequest, this, &Private::out_aboutToSendRequest);
+				aboutToSendRequestConnection[woh] = woh->aboutToSendRequest.connect(boost::bind(&Private::out_aboutToSendRequest, this, woh));
```
jkarneges, line 534: I believe there is just one of these for the session, so map not needed.

src/cpp/proxy/websocketoverhttp.cpp
sima-fastly: wow! nice catch!

jkarneges (approved):

PR #47842, opened 2023-12-14
rename condure to pushpin-condure

jkarneges:
To avoid conflicts with any separately-installed condure.

For v2 we'll probably rename it again to something like pushpin-connmgr, but either way we'll want this prefixing.

### DIFF
--- a/examples/config/pushpin.conf
+++ b/examples/config/pushpin.conf
@@ -19,7 +19,7 @@ stats_connection_send=true
 
 [runner]
 # services to start
-services=condure,zurl,pushpin-proxy,pushpin-handler
+services=condure,pushpin-proxy,pushpin-handler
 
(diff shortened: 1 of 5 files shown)


examples/config/pushpin.conf
```diff
@@ -19,7 +19,7 @@ stats_connection_send=true
 
 [runner]
 # services to start
-services=condure,zurl,pushpin-proxy,pushpin-handler
```
maddie-boby, line 22: Is this intentional in this commit?

jkarneges: Ah yes, I forgot to mention: since we will always prefer our bundled condure, we can assume it has all the features we need, including its outbound connection support which supersedes zurl.

sima-fastly (approved):

PR #47846, opened 2023-12-15
boostification of process quit

sima-fastly:
locally tested manually

### DIFF
--- a/src/cpp/processquit.cpp
+++ b/src/cpp/processquit.cpp
@@ -53,12 +53,14 @@ class SafeSocketNotifier : public QObject
 {
 	Q_OBJECT
 public:
+	Connection activatedConnection;
+
 	SafeSocketNotifier(int socket, QSocketNotifier::Type type,
 		QObject *parent = 0) :
 		QObject(parent)
 	{
 		sn = new QSocketNotifier(socket, type, this);
-		connect(sn, SIGNAL(activated(int)), SIGNAL(activated(int)));
+		connect(sn, SIGNAL(activated(int)), SLOT(doActivated()));
 	}
 
 	~SafeSocketNotifier()
@@ -73,6 +75,10 @@ public:
 
 public slots:
 	void setEnabled(bool enable)       { sn->setEnabled(enable); }
+
+	void doActivated(){
+		activated(sn->socket());
+	}
 
 public:
 	SignalInt activated;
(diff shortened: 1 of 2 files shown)


src/cpp/processquit.cpp
```diff
@@ -102,6 +102,9 @@ inline bool is_gui_app()
 class ProcessQuit::Private : public QObject
 {
 	Q_OBJECT
+
+	Connection activatedConnection;
+
```
jkarneges, line 107: I'd move this to the public section below. Typically, whenever we declare a "Private" class, all its vars are public.

src/cpp/processquit.cpp
```diff
@@ -74,8 +74,8 @@ class SafeSocketNotifier : public QObject
 public slots:
 	void setEnabled(bool enable)       { sn->setEnabled(enable); }
 
-signals:
-	void activated(int socket);
+public:
+	SignalInt activated;
```
jkarneges, line 78: This change won't work since the connect call's destination is a Qt signal. To make it work, connect to a Qt slot that calls the boost signal.

sima-fastly: I see makes senes! updating

sima-fastly: after another look, i realized the connect destination is in fact a slot. can you please clarify your concern.

jkarneges: Line 61 specifies a signal as the destination.

sima-fastly: got it, updated, thanks

src/cpp/processquit.cpp
```diff
@@ -74,8 +76,12 @@ class SafeSocketNotifier : public QObject
 public slots:
 	void setEnabled(bool enable)       { sn->setEnabled(enable); }
 
-signals:
-	void activated(int socket);
+	void doActivated(){
+		activated(sn->socket());
+	}
+
```
jkarneges, line 82: Let's move this to a `private slots:` section at the end.

Also, opening `{` should be on the next line (unless written compactly like `func() { ... }`, which would be acceptable here too).

src/cpp/processquit.cpp
```diff
@@ -213,6 +221,7 @@ public slots:
 #endif
 	}
 
+public:
 	void sig_activated(int)
 	{
```
jkarneges, line 226: Let's move this function up to the earlier public section rather than having two public sections.

jkarneges (approved):

PR #47853, opened 2024-01-04
handle non-chunked responses of indefinite length

jkarneges:


sima-fastly (approved):
looks great, I would just recommend moving out parts of the logic to its own fn, for instance, chunked transfer handling can be moved to a separate function to be called for readability purposes.

jkarneges:
Good idea, I've moved things around.

PR #47854, opened 2024-01-04
Makefile: force phony targets

jkarneges:


madeline-pratt (approved):

sima-fastly (commented):
if these are phony targets why not do sth like this:
.PHONY: build cargo-test ...

jkarneges:
I noticed the qmake-generated makefiles are doing it this way, which seems to be for [compatibility with non-GNU make](https://www.gnu.org/software/make/manual/html_node/Force-Targets.html). So just some defensive coding. That said, I don't know if our makefiles actually work with any non-GNU makes.

PR #47856, opened 2024-01-04
Use dtolnay/rust-toolchain instead of actions-rs/toolchain

jannic:
Reason: actions-rs/toolchain is no longer supported.


jkarneges:
Resolved in #47859

PR #47859, opened 2024-01-05
Use dtolnay/rust-toolchain instead of actions-rs/toolchain

jkarneges:
Reason: actions-rs/toolchain is no longer supported.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,7 +13,7 @@ jobs:
       with:
         submodules: recursive
     - name: Install Rust
-      uses: actions-rs/toolchain@v1
+      uses: dtolnay/rust-toolchain@master
       with:
         toolchain: ${{ matrix.rust-toolchain }}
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -13,7 +13,7 @@ jobs:
       with:
         submodules: recursive
     - name: Install Rust
-      uses: actions-rs/toolchain@v1
+      uses: dtolnay/rust-toolchain@v1
```
jannic, line 16:
```suggestion
      uses: dtolnay/rust-toolchain@master
```
The tag `v1` is not supported. (I already commented on the commit, I wasn't aware that such a comment wouldn't show up on the pull request automatically.)
https://github.com/fastly/pushpin/commit/2820d3da5138ea86a9351aef297cd487cc961bf1#r136287153

madeline-pratt (approved):

PR #47862, opened 2024-01-05
fix more clippy warnings

jkarneges:


src/connection.rs
```diff
@@ -994,7 +994,7 @@ impl<'a, R: AsyncRead, W: AsyncWrite> RequestStartResponse<'a, R, W> {
             if let Err(e) = recv_nonzero(&mut self.r.stream, self.r.buf1).await {
                 if e.kind() == io::ErrorKind::WriteZero {
                     // if there's no more space, suspend forever
-                    let () = std::future::pending().await;
+                    std::future::pending::<()>().await;
```
jannic, line 997: Alternatively, you could just disable the clippy warning. I think the old code was perfectly reasonable, and the clippy warning was a false positive.
But the new code is also fine, albeit less readable.

jannic (approved):

PR #47866, opened 2024-01-05
boostification of proxy session

sima-fastly:
boostification of proxy session

tested manually

### DIFF
--- a/src/cpp/proxy/engine.cpp
+++ b/src/cpp/proxy/engine.cpp
@@ -782,7 +782,9 @@ private slots:
 		ProxyItem *i = proxyItemsBySession.value(ps);
 		assert(i);
 
+		addNotAllowedConnection.erase(ps);
 		finishedConnection.erase(ps);
+		reqSessionDestroyedConnection.erase(ps);
 		
 		if(i->shared)
 			proxyItemsByKey.remove(i->key);
--- a/src/cpp/proxy/engine.cpp
+++ b/src/cpp/proxy/engine.cpp
@@ -116,9 +116,9 @@ public:
 	Connection requestReadyConnection;
 	Connection socketReadyConnection;
 	Connection iRequestReadyConnection;
-	Connection addNotAllowedConnection;
-	Connection finishedConnection;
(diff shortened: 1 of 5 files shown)


src/cpp/proxy/proxysession.h
```diff
@@ -65,10 +66,9 @@ class ProxySession : public QObject
 	// takes ownership
 	void add(RequestSession *rs);
 
-signals:
-	void addNotAllowed(); // no more sharing, for whatever reason
-	void finished();
-	void requestSessionDestroyed(RequestSession *rs, bool accept);
+	Signal addNotAllowed; // no more sharing, for whatever reason
+	Signal finished;
+    boost::signals2::signal<void(RequestSession*, bool)> requestSessionDestroyed;
```
jkarneges, line 71: Indentation

src/cpp/proxy/engine.cpp
```diff
@@ -116,6 +116,9 @@ class Engine::Private : public QObject
 	Connection requestReadyConnection;
 	Connection socketReadyConnection;
 	Connection iRequestReadyConnection;
+	Connection addNotAllowedConnection;
+	Connection finishedConnection;
+	Connection reqSessionDestroyedConnection;
```
jkarneges, line 121: There could be more than one of these.

src/cpp/proxy/engine.cpp
```diff
@@ -776,13 +777,13 @@ private slots:
...
 
-	void ps_finished()
+	void ps_finished(ProxySession *ps)
 	{
-		ProxySession *ps = (ProxySession *)sender();
-
 		ProxyItem *i = proxyItemsBySession.value(ps);
 		assert(i);
 
+		finishedConnection.erase(ps);
```
jkarneges, line 785: Need to delete from the other maps too.

sima-fastly: thanks for catching this one.
I also simplified another conn map erase `d->finishedConnections.erase(req);`

jkarneges (approved):

PR #47873, opened 2024-01-10
improve dependency detection

jkarneges:
Support more Qt lib variations, and fix a case where the boost version wasn't checked.

Tested on mac and linux, including linuxbrew, with Qt 5 & 6.


sima-fastly (approved):
looks great, thanks for the updates and fixes

PR #47875, opened 2024-01-12
boostification of handlerengine

sima-fastly:
tested manually

### DIFF
--- a/src/cpp/handler/handlerengine.cpp
+++ b/src/cpp/handler/handlerengine.cpp
@@ -1170,8 +1170,7 @@ public:
 		timer_->start(SUBSCRIBED_DELAY);
 	}
 
-signals:
-	void subscribed();
+	Signal subscribed;
 
 private:
 	QString channel_;
@@ -1180,7 +1179,7 @@ private:
 private slots:
 	void timer_timeout()
 	{
-		emit subscribed();
+		subscribed();
 	}
 };
 
@@ -1252,6 +1251,7 @@ public:
 	Connection controlServerConnection;
 	Connection itemReadyConnection;
 	map<Deferred*, Connection> finishedConnection;
+	map<Subscription*, Connection> subscribedConnection;
 
 	Private(HandlerEngine *_q) :
 		QObject(_q),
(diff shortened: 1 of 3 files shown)


src/cpp/handler/handlerengine.cpp
```diff
@@ -818,9 +818,8 @@ class AcceptWorker : public Deferred
 		return out;
 	}
 
-signals:
-	void sessionsReady();
-	void retryPacketReady(const RetryRequestPacket &packet);
+	Signal sessionsReady;
+    boost::signals2::signal<void(const RetryRequestPacket&)> retryPacketReady;
```
jkarneges, line 822: indentation

src/cpp/handler/handlerengine.cpp
```diff
@@ -1251,7 +1249,14 @@ class HandlerEngine::Private : public QObject
...
 	Connection itemReadyConnection;
-	map<Deferred*, Connection> finishedConnection;
+	map<Deferred*, Connection> finishedDConnection;
+	map<RefreshWorker*, Connection> finishedRWConnection;
+	map<AcceptWorker*, Connection> finishedAWConnection;
+	map<InspectWorker*, Connection> finishedIWConnection;
+	map<ConnCheckWorker*, Connection> finishedCCWConnection;
+	map<Subscription*, Connection> subscribedConnection;
+	map<AcceptWorker*, Connection> retryPacketReadyConnection;
+	map<AcceptWorker*, Connection> sessionsReadyConnection;
```
jkarneges, line 1259: RefreshWorker and ConnCheckWorker both inherit Deferred, so I think they could stay using the `finishedConnection` map?

For AcceptWorker, this could be an opportunity to try a map of a list.

jkarneges (approved):

src/cpp/handler/handlerengine.cpp
sima-fastly: ah thats why did work in the first place! 
all thoes workers including `AcceptWorker ` inherit `Deferred` . all replaced

PR #47878, opened 2024-01-16
handler: allow configuring multiple proxy accept/stats specs

jkarneges:
This allows one handler to receive stats from multiple proxies. Proxy instances send most stats types over the stats socket, but they also send conn-max over the accept socket, so the handler needs to be able to connect to multiple of both.

With this change, it is possible to then tell the handler to connect to multiple proxy instances like this:

```
[handler]
...
proxy_accept_specs=ipc://{rundir}/{ipc_prefix}accept-1,ipc://{rundir}/{ipc_prefix}accept-2
proxy_stats_specs=ipc://{rundir}/{ipc_prefix}proxy-stats-1,ipc://{rundir}/{ipc_prefix}proxy-stats-2
```

Notes:

* Running multiple handlers (e.g. N handlers and 1 proxy, or N handlers and M proxies) will result in incorrect stats. This change only enables stats to work with 1 handler and N proxies.
* Other things are broken when running multiple proxies, such as retry and websocket control sessions. This change is just a first step towards eventually getting multiple proxies to work.

### DIFF
--- a/src/cpp/handler/handlerapp.cpp
+++ b/src/cpp/handler/handlerapp.cpp
@@ -248,14 +248,22 @@ public:
 		QStringList intreq_in_specs = settings.value("handler/proxy_intreq_in_specs").toStringList();
 		trimlist(&intreq_in_specs);
 		QString proxy_inspect_spec = settings.value("handler/proxy_inspect_spec").toString();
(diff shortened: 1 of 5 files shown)


sima-fastly (approved):
great first step

PR #47880, opened 2024-01-19
remove condure_bin setting, since the internal condure is always used

jkarneges:


sima-fastly (approved):
would be nice to have zurl out at some point as well

PR #47882, opened 2024-01-19
zhttpsocket: proper handling of is_writable

jkarneges:
`AsyncSender::is_writable` and `wait_writable` indicate if/when the channel might be writable, but they do not guarantee the channel is actually writable. This means if `wait_writable().await` is followed by a non-blocking send, the possibility of the send failing must still be handled. This change ensures failed sends are reattempted when writability is indicated again.

This change also corrects `check_send_any` to wait for at least one handle to be writable instead of waiting for all to be writable, and it clarifies the various sending error scenarios.

### DIFF
--- a/src/zhttpsocket.rs
+++ b/src/zhttpsocket.rs
@@ -272,6 +272,11 @@ struct SessionItem {
     handle_index: usize,
 }
 
+enum SessionAddError {
+    Full,
+    Exists,
+}
+
 struct SessionDataInner {
     items: Slab<SessionItem>,
     items_by_key: HashMap<SessionKey, usize>,
@@ -292,11 +297,15 @@ impl SessionData {
         }
     }
 
-    fn add(&self, key: SessionKey, handle_index: usize) -> Result<Session, ()> {
+    fn add(&self, key: SessionKey, handle_index: usize) -> Result<Session, SessionAddError> {
         let inner = &mut *self.inner.lock().unwrap();
 
-        if inner.items.len() == inner.items.capacity() || inner.items_by_key.contains_key(&key) {
-            return Err(());
+        if inner.items.len() == inner.items.capacity() {
+            return Err(SessionAddError::Full);
+        }
+
+        if inner.items_by_key.contains_key(&key) {
+            return Err(SessionAddError::Exists);
         }
 
         let item_key = inner.items.insert(SessionItem {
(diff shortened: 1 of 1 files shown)


src/zhttpsocket.rs
```diff
@@ -292,11 +297,15 @@ impl SessionData {
         }
     }
 
-    fn add(&self, key: SessionKey, handle_index: usize) -> Result<Session, ()> {
+    fn add(&self, key: SessionKey, handle_index: usize) -> Result<Session, SessionAddError> {
         let inner = &mut *self.inner.lock().unwrap();
```
sima-fastly, line 301: mutex handling can be improved improved by handling the potential error that can occur when locking the mutex, instead of using unwrap()

sima-fastly (commented):
there are a few places in this file where we see unwrap() instead of err handling such as 
```
            // can't fail
            let (frame, _) = tnetstring::parse_frame(mi.data).unwrap();
```
do you think adding some log lines in these cases would be helpful instead of panic ?

sima-fastly (commented):
overall looks great 🎖️  
posted some comments and Qs

src/zhttpsocket.rs
```diff
@@ -2268,34 +2345,47 @@ impl ServerSocketManager {
...
+            };
+
+            handles.send_any(&msg, from, ids)
         };
 
-        handles.send_any(&msg, from, ids);
+        match ret {
+            Ok(()) => {}
+            Err(StreamHandlesSendError::BadFormat) => warn!("stream send_any: bad format"),
+            Err(StreamHandlesSendError::NoneReady) => *next_msg = Some(msg),
```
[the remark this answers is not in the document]
jkarneges: yup

src/zhttpsocket.rs
jkarneges: `lock()` fails if the mutex is poisoned, which means we have a logic error and so it's reasonable to abort the program on the spot

jkarneges:
> there are a few places in this file where we see unwrap() instead of err handling such as
> 
> ```
>             // can't fail
>             let (frame, _) = tnetstring::parse_frame(mi.data).unwrap();
> ```
> 
> do you think adding some log lines in these cases would be helpful instead of panic ?

this is another case where an error would be a logic error, hence the comment, and panicking is thus preferred

sima-fastly (approved):

PR #47894, opened 2024-01-26
boostificaiton fixes

sima-fastly:
tested manually

### DIFF
--- a/src/cpp/proxy/sockjsmanager.cpp
+++ b/src/cpp/proxy/sockjsmanager.cpp
@@ -68,6 +68,12 @@ static QByteArray serializeJsonString(const QString &s)
 	return tmp.mid(1, tmp.length() - 2);
 }
 
+struct ZhttpReqConnections{
+	Connection readyReadConnection;
+	Connection bytesWrittenConnection;
+	Connection errorConnection;
+};
+
 class SockJsManager::Private : public QObject
 {
 	Q_OBJECT
@@ -133,9 +139,7 @@ public:
 	QByteArray iframeHtml;
 	QByteArray iframeHtmlEtag;
 	QSet<ZhttpRequest*> discardedRequests;
-	Connection readyReadConnection;
-	Connection bytesWrittenConnection;
-	Connection errorConnection;
+	map<ZhttpRequest*, ZhttpReqConnections> reqConnectionMap;
 
 	Private(SockJsManager *_q, const QString &sockJsUrl) :
 		QObject(_q),
@@ -241,9 +245,11 @@ public:
 
 		s->route = route;
 
-		readyReadConnection = req->readyRead.connect(boost::bind(&Private::req_readyRead, this, req));
-		bytesWrittenConnection = req->bytesWritten.connect(boost::bind(&Private::req_bytesWritten, this, boost::placeholders::_1, req));
-		errorConnection = req->error.connect(boost::bind(&Private::req_error, this, req));
+		reqConnectionMap[req] = {
+			req->readyRead.connect(boost::bind(&Private::req_readyRead, this, req)),
+			req->bytesWritten.connect(boost::bind(&Private::req_bytesWritten, this, boost::placeholders::_1, req)),
+			req->error.connect(boost::bind(&Private::req_error, this, req))
+		};
 
 		sessions += s;
 		sessionsByRequest.insert(s->req, s);
@@ -364,9 +370,11 @@ public:
 		{
 			discardedRequests += req;
 
-			readyReadConnection = req->readyRead.connect(boost::bind(&Private::req_readyRead, this, req));
-			bytesWrittenConnection = req->bytesWritten.connect(boost::bind(&Private::req_bytesWritten, this, boost::placeholders::_1, req));
-			errorConnection = req->error.connect(boost::bind(&Private::req_error, this, req));
+			reqConnectionMap[req] = {
+				req->readyRead.connect(boost::bind(&Private::req_readyRead, this, req)),
+				req->bytesWritten.connect(boost::bind(&Private::req_bytesWritten, this, boost::placeholders::_1, req)),
+				req->error.connect(boost::bind(&Private::req_error, this, req))
+			};
 		}
 
 		HttpHeaders headers;
@@ -596,6 +604,7 @@ public:
 			if(req->isFinished())
 			{
 				discardedRequests.remove(req);
+				reqConnectionMap.erase(req);
 				delete req;
 			}
 
@@ -614,6 +623,7 @@ public:
 
 	void req_error(ZhttpRequest *req)
 	{
+		reqConnectionMap.erase(req);
 		if(discardedRequests.contains(req))
 		{
 			discardedRequests.remove(req);
--- a/src/cpp/proxy/sockjsmanager.cpp
+++ b/src/cpp/proxy/sockjsmanager.cpp
@@ -120,8 +120,8 @@ public:
 		{
 			if(req)
 			{
-			    owner->reqConnectionMap.erase(req);
-			    delete req;
+				owner->reqConnectionMap.erase(req);
+				delete req;
 			}
 			delete sock;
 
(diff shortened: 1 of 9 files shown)


src/cpp/proxy/sockjsmanager.cpp
```diff
@@ -241,9 +241,9 @@ class SockJsManager::Private : public QObject
 
 		s->route = route;
 
-		readyReadConnection = req->readyRead.connect(boost::bind(&Private::req_readyRead, this));
-		bytesWrittenConnection = req->bytesWritten.connect(boost::bind(&Private::req_bytesWritten, this, boost::placeholders::_1));
-		errorConnection = req->error.connect(boost::bind(&Private::req_error, this));
+		readyReadConnection = req->readyRead.connect(boost::bind(&Private::req_readyRead, this, req));
+		bytesWrittenConnection = req->bytesWritten.connect(boost::bind(&Private::req_bytesWritten, this, boost::placeholders::_1, req));
+		errorConnection = req->error.connect(boost::bind(&Private::req_error, this, req));
```
jkarneges, line 246: There can be multiple requests at once.

sima-fastly: got it, thanks. will update to a connection map.
to be sure I am not missing another other connections, could we have the same issue here:
https://github.com/fastly/pushpin/blob/0cca88e269056bd930f89c32cbf253e0a82724b6/src/cpp/proxy/updater.cpp#L122
or here: 
https://github.com/fastly/pushpin/blob/0cca88e269056bd930f89c32cbf253e0a82724b6/src/cpp/proxy/websocketoverhttp.cpp#L614

jkarneges: those are fine

src/cpp/proxy/sockjsmanager.h
```diff
@@ -28,6 +28,7 @@
 #include <boost/signals2.hpp>
 #include <boost/signals2.hpp>
 
+using std::map;
```
jkarneges, line 31: This could move to the .cpp

src/cpp/qzmq/src/qzmqsocket.cpp
```diff
@@ -601,7 +601,6 @@ class Socket::Private : public QObject
 		}
 	}
 
-public slots:
 	void sn_read_activated()
```
jkarneges, line 604: I'm surprised this didn't need an int arg?

src/cpp/proxy/sockjsmanager.cpp
```diff
@@ -617,10 +621,9 @@ private slots:
 		}
 	}
 
-	void req_error()
+	void req_error(ZhttpRequest *req)
 	{
-		ZhttpRequest *req = (ZhttpRequest *)sender();
-
+		reqConnectionMap.erase(req);
```
jkarneges, line 626: This should probably be moved down to only be called when doing `delete req`.

src/cpp/qzmq/src/qzmqsocket.cpp
sima-fastly: this signal is not boostified yet. i need to revert this line

src/cpp/proxy/sockjsmanager.cpp
sima-fastly: the reason I added map erase here is that I saw the logic here:
```
		if(discardedRequests.contains(req))
		{
			discardedRequests.remove(req);
			delete req;
			return;
		}
```
and I thought map needs an erase regardless of req existing in discardedRequests or not ... later on in the func we have 
```		Session *s = sessionsByRequest.value(req);
		assert(s);

		if(s->pending)
			s->req = 0;
		else
			removeSession(s);
```

src/cpp/proxy/sockjsmanager.cpp
sima-fastly: @jkarneges your thoughts are appreciated here

jkarneges: https://github.com/fastly/pushpin/pull/47894#discussion_r1467980837 :)

src/cpp/proxy/sockjsmanager.cpp
jkarneges: the code is really confusing but I think disassociating a request from a session, and even removing the session, shouldn't destroy the request. at least this seems to be something that happens in other places. such a request would presumably be destroyed later on somehow

jkarneges: notably, if the request was supposed to remain alive with its signals disconnected, there would have been a qt-style disconnect call here

jkarneges: ah, destroying the session does destroy the request, so that case will need to be handled. I suggest disconnecting the signals from `~Session` via `owner` if possible, else from `removeSession`.

jkarneges: and only do this if the `req` member is non-null

src/cpp/proxy/sockjsmanager.cpp
sima-fastly: somehow I was not seeing that earlier message until now! thanks justin for your patience :)

src/cpp/proxy/sockjsmanager.cpp
sima-fastly: got it so I added `reqConnectionMap.erase(req);` right next to where ever there is a `delete req` and also to `removeSession`

src/cpp/proxy/sockjsmanager.cpp
```diff
@@ -160,10 +167,13 @@ class SockJsManager::Private : public QObject
 		// can't remove unless unlinked
 		assert(!s->ext);
 
+
```
jkarneges, line 170: stray space

src/cpp/proxy/sockjsmanager.cpp
```diff
@@ -110,6 +118,7 @@ class SockJsManager::Private : public QObject
 
 		~Session()
 		{
+			owner->discardedRequests.remove(req);
 			delete req;
```
jkarneges, line 122: I think this should be:
```
if(req)
{
    owner->reqConnectionMap.erase(req);
    delete req;
}
```

src/cpp/proxy/sockjsmanager.cpp
```diff
@@ -160,10 +167,13 @@ class SockJsManager::Private : public QObject
 		// can't remove unless unlinked
 		assert(!s->ext);
 
+
 		// note: this method assumes the session has already been removed
 		//   from pendingSessions if needed
-		if(s->req)
+		if(s->req){
+			reqConnectionMap.erase(s->req);
```
jkarneges, line 174: don't need to remove from the map here if the Session destructor is doing this

src/cpp/proxy/sockjsmanager.cpp
```diff
@@ -110,7 +118,10 @@ class SockJsManager::Private : public QObject
 
 		~Session()
 		{
-			delete req;
+			if req{
+				owner->discardedRequests.remove(req);
```
jkarneges, line 122:
```
owner->reqConnectionMap.erase(req);
```

src/cpp/proxy/sockjsmanager.cpp
```diff
@@ -118,9 +118,10 @@ class SockJsManager::Private : public QObject
 
 		~Session()
 		{
-			if req{
-				owner->discardedRequests.remove(req);
-				delete req;
+			if(req)
+			{
+			    owner->reqConnectionMap.erase(req);
+			    delete req;
```
jkarneges, line 124: indentation :)

jkarneges (approved):

src/cpp/proxy/sockjsmanager.cpp
sima-fastly: darn copy & paste! fixed thanks

PR #47895, opened 2024-01-26
include from id on accept requests

jkarneges:
This lets handler know which proxy sent an accept request. We need this in order to implement multiple retry-seq values, which will be needed to make stats work correctly when retries are used with multiple proxies.

```
[DEBUG] 2024-01-26 09:56:14.923 [handler] accepting 1 requests from pushpin-proxy_32027
```

### DIFF
--- a/src/cpp/handler/handlerengine.cpp
+++ b/src/cpp/handler/handlerengine.cpp
@@ -1,6 +1,6 @@
 /*
  * Copyright (C) 2015-2023 Fanout, Inc.
- * Copyright (C) 2023 Fastly, Inc.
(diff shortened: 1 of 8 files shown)


src/cpp/handler/handlerengine.cpp
```diff
@@ -963,11 +963,13 @@ class AcceptWorker : public Deferred
 			return;
 		}
 
+		QByteArray reqFrom = req->from();
```
sima-fastly, line 966: love the Qt cleanup ;)

sima-fastly (approved):
nice update, 1 step close to multi proxy pushpin

PR #47898, opened 2024-01-30
use correct route and retry-seq for conn-max packets

jkarneges:
Route IDs in stats packets should use the value from the `statsRoute()` method rather than the route's `id` field directly. These are often the same value, but not always. Also, retry-seq values should be tracked by `statsRoute()` as well.

### DIFF
--- a/src/cpp/proxy/engine.cpp
+++ b/src/cpp/proxy/engine.cpp
@@ -867,9 +867,6 @@ private:
 
 		log_debug("IN (retry) %s %s", qPrintable(p.requestData.method), p.requestData.uri.toEncoded().data());
 
-		if(p.retrySeq >= 0)
-			stats->setRetrySeq(p.route, p.retrySeq);
-
 		InspectData idata;
 		if(p.haveInspectInfo)
 		{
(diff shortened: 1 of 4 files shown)


jkarneges:
Background:

Each route (as in a route from the routes file) has an ID, either explicitly provided with an `id=` param, or generated internally (with values like `*,*,*,*`, intended to ensure uniqueness). Stats are aggregated per route ID, except for generated route IDs in which case the stats for all such generated route IDs are aggregated. From the user's perspective, this means there are stats per specified ID, and then stats for everything else. The code sometimes refers to these "everything else" stats as "global stats", but this is misleading as they are not *all* stats, just the stats that aren't under a user-specified ID.

These "everything else" stats are internally tracked in the StatsManager using an empty string as the route ID, and they are externally represented without an ID (their stats packets have no `route` field). As a result, when a user specifies an ID on a route, the stats packets for that route will have a `route` field with the same value, and the user doesn't specify an ID on a route, the route will be given an internally generated ID and the stats packets for that route will have no `route` field. The different IDs then need to be used in the correct places in the code. The `statsRoute()` method returns the value that should be used for stats packets.

Anything working with StatsManager should be sure to use the statsRoute() value, and this PR fixes some places where this wasn't the case (when calling the `getConnMaxPacket` and `setRetrySeq` methods of StatsManager).

src/cpp/proxy/engine.cpp
```diff
@@ -867,9 +867,6 @@ private slots:
 
 		log_debug("IN (retry) %s %s", qPrintable(p.requestData.method), p.requestData.uri.toEncoded().data());
 
-		if(p.retrySeq >= 0)
```
sima-fastly, line 870: Why are we removing stats update here, wouldn’t this mean stats would be outdated?

sima-fastly (approved):
Looks good overall 
One concern I have is around this line of code
```
wsControl->start(route.id, route.separateStats, channelPrefix, inSock->requestUri());
```
Wouldn’t it be better to update route.Id to route.statsRoute here ?

src/cpp/proxy/engine.cpp
jkarneges: these lines were moved to requestsession.cpp, where we're able to call `statsRoute()`

jkarneges: specifically, inside RequestSession::startRetry, which is called below

sima-fastly: got it thanks for clarifying

PR #47899, opened 2024-01-31
boostification of all websocket signals

sima-fastly:
boostification of all websocket signals

tested manually

### DIFF
--- a/src/cpp/proxy/sockjssession.cpp
+++ b/src/cpp/proxy/sockjssession.cpp
@@ -160,7 +160,7 @@ public:
 	bool updating;
 	Connection bytesWrittenConnection;
 	Connection errorConnection;
-	WSConnections* wsConnection;
+	WSConnections wsConnection;
 
 	Private(SockJsSession *_q) :
 		QObject(_q),
@@ -244,7 +244,7 @@ public:
 		}
 		requests.clear();
 
-		delete wsConnection;
+		wsConnection = WSConnections();
 		delete sock;
 		sock = 0;
 
@@ -275,7 +275,7 @@ public:
 		}
 		else
 		{
-			wsConnection = new WSConnections{
+			wsConnection = WSConnections{
 				sock->readyRead.connect(boost::bind(&Private::sock_readyRead, this)),
 				sock->framesWritten.connect(boost::bind(&Private::sock_framesWritten, this, boost::placeholders::_1, boost::placeholders::_2)),
 				sock->writeBytesChanged.connect(boost::bind(&Private::sock_writeBytesChanged, this)),
--- a/src/cpp/proxy/sockjssession.cpp
+++ b/src/cpp/proxy/sockjssession.cpp
@@ -160,7 +160,7 @@ public:
 	bool updating;
 	Connection bytesWrittenConnection;
 	Connection errorConnection;
-	WSConnections wsConnection;
+	WSConnections* wsConnection;
 
 	Private(SockJsSession *_q) :
 		QObject(_q),
@@ -244,6 +244,7 @@ public:
 		}
 		requests.clear();
 
+		delete wsConnection;
 		delete sock;
 		sock = 0;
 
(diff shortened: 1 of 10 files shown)


src/cpp/proxy/wsproxysession.cpp
jkarneges, line 1: hmm, indentation changed in the whole file causing a huge diff

sima-fastly: sheesh! my git client doesnt show the diff when the tabs turn into space and I keep missing em prior to PR.
Ill clean this up and resubmit

sima-fastly: I also need to update my VS code indentation settings so it doesn't convert on save or sth

src/cpp/proxy/websocketoverhttp.cpp
```diff
@@ -46,6 +46,12 @@
 
 namespace {
 
+struct WSConnections {
+    Connection disconnectedConnection;
+    Connection closedConnection;
+    Connection errorConnection;
+};
+
```
jkarneges, line 54: indentation

src/cpp/proxy/wsproxysession.cpp
```diff
@@ -291,6 +309,8 @@ class WsProxySession::Private : public QObject
 	Connection keepAliveConneciton;
 	Connection aboutToSendRequestConnection;
 	map<WsControlSession*, WSProxyConnections> wsProxyConnectionMap;
+	map<WebSocket*, WSConnections> outWSConnectionMap;
+	map<WebSocket*, InWSConnections> inWSConnectionMap;
```
jkarneges, line 313: There is only ever one in sock and one out sock.

src/cpp/proxy/sockjssession.cpp
```diff
@@ -244,7 +244,6 @@ class SockJsSession::Private : public QObject
 		}
 		requests.clear();
 
-		wsConnectionMap.erase(sock);
```
jkarneges: should probably still disconnect here somehow?

the connections object seems to have a default constructor, so maybe just:
```
wsConnection = WSConnections();
```

src/cpp/proxy/wsproxysession.cpp
```diff
@@ -352,7 +352,6 @@ class WsProxySession::Private : public QObject
 
 		cleanupInSock();
 
-		outWSConnectionMap.erase(outSock);
```
jkarneges: same (and others)

src/cpp/proxy/wsproxysession.cpp
```diff
@@ -372,7 +371,6 @@ class WsProxySession::Private : public QObject
 		if(inSock)
 		{
 			connectionManager->removeConnection(inSock);
-			inWSConnectionMap.erase(inSock);
```
jkarneges: same

sima-fastly (commented):
@jkarneges I have updated the code to delete connection struct instace when a req/sock gets deleted.

but this reminds me there are other places such as this file : 
https://github.com/fastly/pushpin/blob/adf96ae5692c8c45e6835b40ea4dfb51f588142f/src/cpp/proxy/requestsession.cpp#L238C4-L238C23
where a req or socket are getting deleted without the corresponding connection getting disconnected. Am I correct in thinking these also need a disconnect statement fix?

jkarneges:
Ah yeah, anywhere an object is deleted that might have had connections, especially if the deletion is done from somewhere that isn't the destructor and the object pointer is also nulled out, any connections should also be disconnected.

I suggest keeping the connections object as a value and not a pointer, or else you'll need to null out the connections object too. The connection object is initially implicitly constructed using its default constructor, so explicitly calling the default constructor later on should return it to its initial state.

sima-fastly:
> Ah yeah, anywhere an object is deleted that might have had connections, especially if the deletion is done from somewhere that isn't the destructor and the object pointer is also nulled out, any connections should also be disconnected.
> 
> I suggest keeping the connections object as a value and not a pointer, or else you'll need to null out the connections object too. The connection object is initially implicitly constructed using its default constructor, so explicitly calling the default constructor later on should return it to its initial state.

I have addressed this and will have a follow up PR to clean these up in other files

jkarneges (approved):

PR #47901, opened 2024-02-01
disconnect when obj is getting deleted

sima-fastly:


src/cpp/proxy/sockjssession.cpp
```diff
@@ -224,7 +229,11 @@ class SockJsSession::Private : public QObject
 		{
 			RequestItem *ri = requests.value(req);
 			assert(ri);
-
+			if(ri->req)
```
jkarneges, line 232: `ri->req` is always non-null, so this check is not needed

src/cpp/proxy/sockjssession.cpp
```diff
@@ -197,6 +197,11 @@ class SockJsSession::Private : public QObject
...
 	void removeRequestItem(RequestItem *ri)
 	{
+		if(ri->req)
+		{
+			bytesWrittenConnection.disconnect();
+			errorConnection.disconnect();
+		}
 		requests.remove(ri->req);
 		delete ri;
 	}
```
jkarneges, line 207: `ri->req` is always non-null, so this check is not needed

jkarneges (approved):

PR #47902, opened 2024-02-02
call invokeMethod on the correct objects

jkarneges:


sima-fastly (approved):
thanks for the cleanup

PR #47908, opened 2024-02-06
invoke refresh actions via wscontrol instead of rpc

jkarneges:
Currently, refresh actions are relayed to the proxy via "command" RPC calls, however the handler->proxy RPC interface doesn't work correctly when there are multiple proxies. This PR changes refresh actions to relay via wscontrol messages instead, which does work with multiple proxies.

Note that refresh RPC calls made to the handler are still relayed to the proxy via RPC. It is only the refresh "publish action" that goes over wscontrol. Also, this PR does nothing to make other RPC calls work with multiple proxies by some other means. It only makes refresh actions work with multiple proxies.

Logs of refresh messages getting sent to multiple proxies:
```
[INFO] 2024-02-05 17:25:00.874 [handler] control: POST /publish/ code=200 10 items=1
[DEBUG] 2024-02-05 17:25:00.874 [handler] relaying to 2 ws-message subscribers
[DEBUG] 2024-02-05 17:25:00.874 [handler] stats: OUT message { "channel": "test", "count": 2, "from": "pushpin-handler_25050", "transport": "ws-message" }
[INFO] 2024-02-05 17:25:00.874 [handler] publish channel=test receivers=2
[DEBUG] 2024-02-05 17:25:00.874 [handler] OUT wscontrol: to=pushpin-proxy_25029 { "items": [ { "type": "refresh", "cid": "07276246-d68b-4e75-b2c7-63d0784f02d5" } ], "from": "pushpin-handler_25050" }
[DEBUG] 2024-02-05 17:25:00.874 [handler] OUT wscontrol: to=pushpin-proxy_25026 { "items": [ { "type": "refresh", "cid": "9967abaf-5219-417d-8816-ba46bfbc15de" } ], "from": "pushpin-handler_25050" }
```

### DIFF
--- a/src/cpp/handler/handlerengine.cpp
+++ b/src/cpp/handler/handlerengine.cpp
@@ -1831,10 +1831,7 @@ private:
 			}
 			else if(f.action == PublishFormat::Refresh)
 			{
-				Deferred *d = ControlRequest::refresh(proxyControlClient, i.cid, this);
-				finishedConnection[d] = d->finished.connect(boost::bind(&Private::deferred_finished, this, boost::placeholders::_1, d));
-				deferreds += d;
-				return;
+				i.type = WsControlPacket::Item::Refresh;
 			}
 
 			writeWsControlItems(s->peer, QList<WsControlPacket::Item>() << i);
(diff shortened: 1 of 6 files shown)


sima-fastly (commented):
looks good overall.
Do we need to add a check statement here https://github.com/fastly/pushpin/blob/main/src/cpp/handler/wscontrolmessage.cpp#L53

jkarneges:
Great question! But, no. :) The type names are confusing. WsControlPacket is the class used for messages sent over the wscontrol ipc, and so the new item type was added there. WsControlMessage is something else, used for GRIP control messages received from the backend.

sima-fastly (approved):

PR #47909, opened 2024-02-06
boostification of req session finished signal

sima-fastly:
final boostification PR


src/cpp/proxy/engine.cpp
```diff
@@ -450,6 +451,7 @@ class Engine::Private : public QObject
 
 		// proxysession will take it from here
 		// TODO: use callbacks for performance
+		reqSessionConnectionMap.erase(rs);
 		rs->disconnect(this);
```
jkarneges, line 455: If there are no more qt connections between engine and RequestSession then this line can be removed.

jkarneges (approved):

PR #47915, opened 2024-02-09
nicer build failure output

jkarneges:
When a command fails, exit with an error string instead of panicking, to reduce output noise. For added clarity, output attempted commands to stdout, and when capturing the stdout of a command don't capture its stderr.


sima-fastly (approved):
so needed. thanks for this update 🎖️ 

The initial part of both check_command and check_command_capture functions, which constructs the program and args strings, is repeated. we can refactor these into a separate function to reduce code duplication.

jkarneges:
ah yes, agree about the duplicated parts. will think about how to make that better

PR #47916, opened 2024-02-09
fix handler warnings

jkarneges:


src/cpp/handler/variantutil.cpp
```diff
@@ -271,8 +274,8 @@ bool convertToJsonStyleInPlace(QVariant *in)
 
 	bool changed = false;
 
-	int type = in->type();
-	if(type == QVariant::Hash)
+	int type = typeId(*in);
```
mhp, line 277: Should this be `QMetaType::Type`, not `int`?  I think you've made that substitution elsewhere.

mhp (approved):
LGTM.  One comment about a possible change of type.

PR #47918, opened 2024-02-10
deny c++ warnings when RUSTFLAGS contains "-D warnings"

jkarneges:
CI already sets RUSTFLAGS in this manner, so this change will immediately cause C++ warnings to be denied in CI runs, including the run for this PR.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -110,6 +110,7 @@ fn write_cpp_conf_pri(
     dest: &Path,
     release: bool,
     include_paths: &[&Path],
+    deny_warnings: bool,
 ) -> Result<(), Box<dyn Error>> {
     let mut f = fs::File::create(dest)?;
 
@@ -125,6 +126,12 @@ fn write_cpp_conf_pri(
 
     for path in include_paths {
         writeln!(&mut f, "INCLUDEPATH += {}", path.display())?;
+    }
+
+    writeln!(&mut f)?;
+
+    if deny_warnings {
+        writeln!(&mut f, "QMAKE_CXXFLAGS += \"-Werror\"")?;
     }
 
     Ok(())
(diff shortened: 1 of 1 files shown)


build.rs
```diff
@@ -390,6 +397,10 @@ fn find_boost_include_dir() -> Result<PathBuf, Box<dyn Error>> {
     .into())
 }
 
+fn contains_subslice<T: PartialEq>(haystack: &[T], needle: &[T]) -> bool {
+    haystack.windows(needle.len()).any(|w| w == needle)
```
silentbicycle, line 401: Wouldn't this match if *any* of "-D" or "warnings" match, rather than both (consecutively)? Or false positives for anything with "-D" in the middle?

This seems unlikely to hit false positives in RUSTFLAGS, but it'd be really confusing if it did, so I want to double check.

silentbicycle (changes requested):
Looks good, but there's a thing I want to double check.

build.rs
jkarneges: `windows(needle.len())` returns a sequence of subslices, and `w == needle` tests the slices for equality which should mean checking they contain the exact same elements in the same order

jkarneges: [windows() in playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=0def39d4a8115f3cbbd540b4ed3c6862)

silentbicycle: Okay -- `needle.len()` refers to consecutive _words_ here rather than the character length of the individual needle elements, so false positives from substring matches aren't an issue.

silentbicycle (approved):
Looks good

jkarneges:
thank you!

PR #47922, opened 2024-02-26
mark various globals as thread local, to enable thread-safe use

jkarneges:
We currently don't use these APIs from multiple threads but after this change we will be able to.


sima-fastly (approved):
lgtm

PR #47924, opened 2024-02-27
proxy: rename workers to sessions, and cap sessions to maxconn

jkarneges:
Currently, the proxy uses the term "workers" to refer to connection/session handlers (for example if maxWorkers is 1000 then up to 1000 concurrent sessions can be supported). This PR changes the term from "workers" to "sessions", to free up the term "workers" so that we can use it for worker threads.


sima-fastly (approved):
lgtm

PR #47926, opened 2024-02-27
proxy: run engine in separate thread

jkarneges:
This runs the engine in a separate thread instead of the main thread, to prepare for supporting multiple threads. It also moves ownership of DomainMap out of the engine, to make it possible for multiple engines to share it.

### DIFF
--- a/src/cpp/proxy/app.cpp
+++ b/src/cpp/proxy/app.cpp
@@ -33,8 +33,11 @@
 #include "log.h"
 #include "settings.h"
 #include "xffrule.h"
+#include "domainmap.h"
 #include "engine.h"
 #include "config.h"
+
+using Connection = boost::signals2::scoped_connection;
 
 static void trimlist(QStringList *list)
 {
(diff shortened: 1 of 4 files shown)


src/cpp/proxy/app.cpp
```diff
@@ -162,21 +165,149 @@ static CommandLineParseResult parseCommandLine(QCommandLineParser *parser, ArgsD
 	return CommandLineOk;
 }
 
+class EngineWorker : public QObject
```
sima-fastly, line 168: can we refrain from adding QObject to our new code logic as we are moving away from this approach all together

jkarneges: fair question! here it is still needed, due to `QMetaObject::invokeMethod`.

sima-fastly: due to QMetaObject::invokeMethod we need to include `Q_OBJECT` macro but I do not think we need the `QObject` as a parent class

sima-fastly (approved):

PR #47927, opened 2024-02-28
removed qobj parent from deferred

sima-fastly:
This PR intends to remove QObject parent param from Deferred class.
We are not removing Q_Object altogether from Deferred yet as `QMetaObject::invokeMethod` is being used in Deferred which we will tackle at a later time.
All new obj creations are done using unique pointers now so we can manage their lifetime properly
We have also removed QObject parent param from other class constructors to remove any errors or warnings.

### DIFF
--- a/src/cpp/handler/conncheckworker.cpp
+++ b/src/cpp/handler/conncheckworker.cpp
@@ -28,8 +28,8 @@
 #include "controlrequest.h"
 #include "statsmanager.h"
 
-ConnCheckWorker::ConnCheckWorker(ZrpcRequest *req, ZrpcManager *proxyControlClient, StatsManager *stats, QObject *parent) :
-	Deferred(parent),
+ConnCheckWorker::ConnCheckWorker(ZrpcRequest *req, ZrpcManager *proxyControlClient, StatsManager *stats) :
+	Deferred(),
 	req_(req)
 {
 	req_->setParent(this);
@@ -64,7 +64,7 @@ ConnCheckWorker::ConnCheckWorker(ZrpcRequest *req, ZrpcManager *proxyControlClie
 	if(!missing_.isEmpty())
 	{
 		// ask the proxy about any cids we don't know about
-		Deferred *d = ControlRequest::connCheck(proxyControlClient, missing_, this);
+		Deferred *d = ControlRequest::connCheck(proxyControlClient, missing_);
 		finishedConnection_ = d->finished.connect(boost::bind(&ConnCheckWorker::proxyConnCheck_finished, this, boost::placeholders::_1));
 		return;
 	}
--- a/src/cpp/handler/conncheckworker.cpp
+++ b/src/cpp/handler/conncheckworker.cpp
@@ -64,7 +64,7 @@ ConnCheckWorker::ConnCheckWorker(ZrpcRequest *req, ZrpcManager *proxyControlClie
 	if(!missing_.isEmpty())
 	{
 		// ask the proxy about any cids we don't know about
-		Deferred *d = ControlRequest::connCheck(proxyControlClient, missing_);
+		auto d = ControlRequest::connCheck(proxyControlClient, missing_);
 		finishedConnection_ = d->finished.connect(boost::bind(&ConnCheckWorker::proxyConnCheck_finished, this, boost::placeholders::_1));
 		return;
 	}
--- a/src/cpp/handler/conncheckworker.cpp
+++ b/src/cpp/handler/conncheckworker.cpp
@@ -64,7 +64,7 @@ ConnCheckWorker::ConnCheckWorker(ZrpcRequest *req, ZrpcManager *proxyControlClie
 	if(!missing_.isEmpty())
 	{
 		// ask the proxy about any cids we don't know about
-		auto d = ControlRequest::connCheck(proxyControlClient, missing_);
+		auto d = std::unique_ptr<Deferred>(ControlRequest::connCheck(proxyControlClient, missing_));
 		finishedConnection_ = d->finished.connect(boost::bind(&ConnCheckWorker::proxyConnCheck_finished, this, boost::placeholders::_1));
 		return;
 	}
--- a/src/cpp/handler/controlrequest.cpp
+++ b/src/cpp/handler/controlrequest.cpp
@@ -34,7 +34,6 @@ class ConnCheck : public Deferred
 {
 	Q_OBJECT
 	
-	std::unique_ptr<ZrpcRequest> req;
 	Connection finishedConnection;
 
 public:
@@ -54,6 +53,8 @@ public:
 	}
 
 private:
+	std::unique_ptr<ZrpcRequest> req;
+
 	void req_finished(ZrpcRequest *req)
 	{
 		if(req->success())
@@ -92,7 +93,6 @@ class Refresh : public Deferred
 {
 	Q_OBJECT
 	
-	std::unique_ptr<ZrpcRequest> req;
 	Connection finishedConnection;
 
 public:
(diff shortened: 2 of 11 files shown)


jkarneges (commented):
The removing of parenting here looks fine, but we lose automatic cleanup of the object tree, so we need some replacement for that.

src/cpp/handler/controlrequest.cpp
```diff
@@ -37,11 +37,11 @@ class ConnCheck : public Deferred
...
 
 public:
-	ConnCheck(ZrpcManager *controlClient, const CidSet &cids, QObject *parent = 0) :
-		Deferred(parent)
+	ConnCheck(ZrpcManager *controlClient, const CidSet &cids) :
+		Deferred()
 	{
-		ZrpcRequest *req = new ZrpcRequest(controlClient, this);
-		finishedConnection = req->finished.connect(boost::bind(&ConnCheck::req_finished, this, req));
+		auto req = std::make_unique<ZrpcRequest>(controlClient, this);
```
jkarneges, line 43: Will need to save this in a member to avoid it getting destroyed at the end of this function.

Should also remove passing `this` as the parent, since its lifetime will be tracked with the `unique_ptr`.

src/cpp/handler/conncheckworker.cpp
```diff
@@ -64,7 +64,7 @@ ConnCheckWorker::ConnCheckWorker(ZrpcRequest *req, ZrpcManager *proxyControlClie
 	if(!missing_.isEmpty())
 	{
 		// ask the proxy about any cids we don't know about
-		Deferred *d = ControlRequest::connCheck(proxyControlClient, missing_, this);
+		Deferred *d = ControlRequest::connCheck(proxyControlClient, missing_);
```
jkarneges, line 67: Need to keep track of this one too, and others.

src/cpp/handler/controlrequest.cpp
```diff
@@ -33,14 +33,15 @@ namespace ControlRequest {
 class ConnCheck : public Deferred
 {
 	Q_OBJECT
-
+	
+	std::unique_ptr<ZrpcRequest> req;
 	Connection finishedConnection;
```
jkarneges, line 38: move these to the `private:` section below?

jkarneges (commented):
Looks like some constructing functions return a unique_ptr, and other times the caller chooses to wrap with unique_ptr. I wonder about making these usages more consistent. How about we always have the caller choose, as that's the most flexible? I think you can use unique_ptr's regular constructor for situations where `make_unique` may not work.

src/cpp/handler/controlrequest.cpp
```diff
@@ -33,15 +33,15 @@ namespace ControlRequest {
...
 
 public:
-	ConnCheck(ZrpcManager *controlClient, const CidSet &cids, QObject *parent = 0) :
-		Deferred(parent)
+	ConnCheck(ZrpcManager *controlClient, const CidSet &cids) :
+		Deferred()
 	{
-		ZrpcRequest *req = new ZrpcRequest(controlClient, this);
-		finishedConnection = req->finished.connect(boost::bind(&ConnCheck::req_finished, this, req));
+		req = std::make_unique<ZrpcRequest>(controlClient, this);
```
jkarneges, line 43: Passing `this` as the parent should be removed here, since the lifetime is already managed by unique_ptr and there is a risk of double-deletion if QObject parenting is used at the same time.

src/cpp/handler/handlerengine.cpp
```diff
@@ -232,8 +232,8 @@ class InspectWorker : public Deferred
 			if(getSession && stateClient)
 			{
 				// determine session info
-				Deferred *d = SessionRequest::detectRulesGet(stateClient, requestData.uri.host().toUtf8(), requestData.uri.path(QUrl::FullyEncoded).toUtf8(), this);
-				finishedConnection[d] = d->finished.connect(boost::bind(&InspectWorker::sessionDetectRulesGet_finished, this, boost::placeholders::_1));
+				auto d = std::unique_ptr<Deferred>(SessionRequest::detectRulesGet(stateClient, requestData.uri.host().toUtf8(), requestData.uri.path(QUrl::FullyEncoded).toUtf8()));
```
jkarneges, line 235: Needs to be a member to stay alive.

src/cpp/handler/handlerengine.cpp
```diff
@@ -1298,11 +1297,12 @@ class HandlerEngine::Private : public QObject
 		proxyStatsSock(0),
 		proxyStatsValve(0),
 		controlHttpServer(0),
-		stats(0),
-		report(0)
+		stats(0)
 	{
 		qRegisterMetaType<DetectRuleList>();
 
+		proxyControlClient = std::make_unique<ZrpcManager>(this);
```
jkarneges, line 1304: this already happens later and probably doesn't need to be here?

relatedly: watch out about that QObject parenting

src/cpp/handler/handlerengine.cpp
```diff
@@ -2474,9 +2474,9 @@ class HandlerEngine::Private : public QObject
 			all.httpResponseMessagesSent += qMax(p.httpResponseMessagesSent, 0);
 		}
 
-		report = ControlRequest::report(proxyControlClient, all, this);
-		finishedConnection[report] = report->finished.connect(boost::bind(&Private::report_finished, this, boost::placeholders::_1));
-		deferreds += report;
+		report = std::unique_ptr<Deferred>(ControlRequest::report(proxyControlClient.get(), all));
+		finishedConnection[report.get()] = report->finished.connect(boost::bind(&Private::report_finished, this, boost::placeholders::_1));
+		deferreds[report.get()] = std::move(report);
```
jkarneges, line 2479: storing the pointer in two places like this doesn't really work with unique_ptr I don't think. I assume the `std::move` nulls the `report`. it looks like `report` is only used in `stats_reported` to guard against starting multiple report rpc calls, and it only checks if it's non-null, so maybe just get rid of `report` and replace it with `bool reporting`.

src/cpp/handler/controlrequest.cpp
sima-fastly: zrprequest is still not updated in this regard yet. 
```
ZrpcRequest(ZrpcManager *manager, QObject *parent = 0);
```

src/cpp/handler/handlerengine.cpp
sima-fastly: `std::move` is to move ownership of that particular unique_ptr to deferreds.
we still need `report` to keep track of `finishedConnection` thats based on report.

src/cpp/handler/handlerengine.cpp
sima-fastly: I think a good way to keep this alive as long as needed is to use deferreds as such 
```
deferreds[d.get()] = std::move(d);
```

sima-fastly: maybe we can use a data structure similar to `deferreds` here
Also looking through the code, connections in InspectWorker and AcceptWorker are not erased from the map. maybe at the end of `doFinish()`

src/cpp/handler/handlerengine.cpp
jkarneges: [unique_ptr's docs](https://en.cppreference.com/w/cpp/memory/unique_ptr/unique_ptr) for its move constructor:

```c++
unique_ptr( unique_ptr&& u ) noexcept;             (5)
```

> 5) Constructs a unique_ptr by transferring ownership from u to *this and stores the null pointer in u.

so the `if(report)` will always return false, and we'll need some other way to track the finished connection

jkarneges: perhaps `shared_ptr` could be used here, since that's logically what the original code was to doing (two handles to the same data), although it would be nice to find a way to not need that

`finishedConnection` might be fine, since it's populated while `report` is still valid, and erase is called with a slot argument instead of `report`

src/cpp/handler/handlerengine.cpp
jkarneges: yeah I think those classes simply clean up when they eventually destruct themselves, compared to the engine which is long lived and needs to regularly clean up as it goes

PR #47930, opened 2024-03-02
proxy: don't use qapp as parent for objects not in main thread

jkarneges:
This was probably done to ensure automatic cleanup on shutdown, but it's not needed since we explicitly call `clearDisconnectManager()`.


sima-fastly (approved):
looks good, if its tested all lgtm

PR #47938, opened 2024-03-05
boost connection disconnect fixes

sima-fastly:
fixing boost connection disconnects

also zrpcReq destroyed signal was fixed


jkarneges:
Looks good, just need to have the ZrpcRequest destructor emit destroyed()

jkarneges (approved):

PR #47940, opened 2024-03-06
use non-zero-sized type for vec to ensure deallocation

jkarneges:
ReusableVec type-erases its inner Vec by casting the element type to `()`, but this seems to optimize away the Vec's destructor. Cast to u8 instead.


sima-fastly (approved):
lgtm

PR #47942, opened 2024-03-07
Removed QObject from RunnerApp

sima-fastly:
Removed QObject from RunnerApp

### DIFF
--- a/src/cpp/runner/runnerapp.cpp
+++ b/src/cpp/runner/runnerapp.cpp
@@ -803,6 +803,8 @@ RunnerApp::RunnerApp() {
     d = std::make_unique<Private>(this);
 }
 
+RunnerApp::~RunnerApp() = default;
+
 void RunnerApp::start()
 {
 	d->start();
--- a/src/cpp/runner/runnerapp.cpp
+++ b/src/cpp/runner/runnerapp.cpp
@@ -803,6 +803,8 @@ RunnerApp::RunnerApp() {
 	d = std::make_unique<Private>(this);
 }
 
+RunnerApp::~RunnerApp() = default;
+
 void RunnerApp::start()
 {
 	d->start();
--- a/src/cpp/runner/runnerapp.cpp
+++ b/src/cpp/runner/runnerapp.cpp
@@ -651,7 +651,7 @@ public:
 				s->error.connect(boost::bind(&Private::service_error, this, boost::placeholders::_1, s))
 			};
 
-			if(!args.mergeOutput)
+			if(!args.mergeOutput || s->alwaysLogStatus())
 				log_info("starting %s", qPrintable(s->name()));
 
 			s->start();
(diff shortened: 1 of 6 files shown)


src/cpp/runner/runnerapp.h
```diff
@@ -46,7 +44,7 @@ class RunnerApp : public QObject
 private:
 	class Private;
 	friend class Private;
-	Private *d;
+	std::unique_ptr<Private> d;
```
jkarneges, line 47: nice

src/cpp/runner/runnerapp.cpp
```diff
@@ -802,20 +799,14 @@ class RunnerApp::Private : public QObject
...
-	d = new Private(this);
+RunnerApp::RunnerApp() {
+    d = std::make_unique<Private>(this);
 }
 
-RunnerApp::~RunnerApp()
-{
-	delete d;
-}
+RunnerApp::~RunnerApp() = default;
```
jkarneges, line 806: I didn't even know you could do that! maybe useful for libs. in our case I'd say we don't need to specify this at all and can just remove the destructor from .cpp & .h when we only want the default.

src/cpp/runner/runnerapp.cpp
```diff
@@ -654,7 +651,7 @@ class RunnerApp::Private : public QObject
 				s->error.connect(boost::bind(&Private::service_error, this, boost::placeholders::_1, s))
 			};
 
-			if(!args.mergeOutput || qobject_cast<Mongrel2Service*>(s))
+			if(!args.mergeOutput)
```
jkarneges, line 654: hmm, guess we broke the casting in an earlier commit. I suggest adding a `bool alwaysLogStatus() const { return false; }` method to `Service` that `Mongrel2Service` overrides to return true, and then call that method here instead of casting.

src/cpp/runner/runnerapp.h
```diff
@@ -31,12 +31,10 @@ using std::map;
...
 
-class RunnerApp : public QObject
+class RunnerApp
 {
-	Q_OBJECT
-
 public:
-	RunnerApp(QObject *parent = 0);
+	RunnerApp();
 	~RunnerApp();
```
jkarneges, line 38: can remove this declaration

jkarneges (approved):

src/cpp/runner/runnerapp.cpp
sima-fastly: we need to keep destructor def due to :
```
function `RunnerAppMain::app_quit(int)':
         runnermain.cpp:41: reference to `RunnerApp::~RunnerApp()'
```

src/cpp/runner/runnerapp.h
sima-fastly: we need to keep destructor def due to :

function `RunnerAppMain::app_quit(int)':
         runnermain.cpp:41: reference to `RunnerApp::~RunnerApp()'

PR #47944, opened 2024-03-07
use immediate mode for router sockets that connect and send

jkarneges:
Immediate mode is needed anytime we use zmq ROUTER sockets that connect and also send (if they bind, or if they connect and never send, this mode is not needed). This mode ensures that peer addresses can be renegotiated correctly anytime a binding peer restarts. Most of our ROUTER sockets already have this set, but I'd forgotten to set it on the ones added recently for multithreading.

Also, don't attempt to send keep alives before peer addr known.

Tested locally.


sima-fastly (approved):
lgtm

PR #47948, opened 2024-03-08
Removed Qobject parent from WsControlManager

sima-fastly:
Removed Qobject parent from WsControlManager


src/cpp/proxy/wscontrolmanager.h
```diff
@@ -26,6 +26,7 @@
 
 #include <QObject>
 #include "packet/wscontrolpacket.h"
+#include <memory>
```
jkarneges, line 29: nit: move above `#include <QObject>`

(our include order is typically: std, lib, local)

jkarneges (approved):

PR #47949, opened 2024-03-08
Removed Qobject Parent from RateLimiter

sima-fastly:
Removed Qobject Parent from RateLimiter

### DIFF
--- a/src/cpp/handler/handlerengine.cpp
+++ b/src/cpp/handler/handlerengine.cpp
@@ -1303,8 +1303,8 @@ public:
 	{
 		qRegisterMetaType<DetectRuleList>();
 
-		publishLimiter = new RateLimiter();
-		updateLimiter = new RateLimiter();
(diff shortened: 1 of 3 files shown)


src/cpp/handler/handlerengine.cpp
```diff
@@ -1303,8 +1303,8 @@ class HandlerEngine::Private : public QObject
 	{
 		qRegisterMetaType<DetectRuleList>();
 
-		publishLimiter = new RateLimiter(this);
-		updateLimiter = new RateLimiter(this);
+		publishLimiter = new RateLimiter;
+		updateLimiter = new RateLimiter;
```
jkarneges, line 1307: these will now need to be cleaned up some other way

sima-fastly: thanks for catching this. working on a fix now

jkarneges (approved):

PR #47950, opened 2024-03-08
removed qobject parent from rtimer

sima-fastly:
removed qobject parent from rtimer


src/cpp/handler/httpsession.cpp
```diff
@@ -213,10 +213,10 @@ class HttpSession::Private : public QObject
 		writeBytesChangedConnection = req->writeBytesChanged.connect(boost::bind(&Private::req_writeBytesChanged, this));
 		errorConnection = req->error.connect(boost::bind(&Private::req_error, this));
 
-		timer = new RTimer(this);
+		timer = new RTimer();
```
jkarneges, line 216: the `()` can be omitted if there are no args

jkarneges (approved):

PR #47958, opened 2024-03-11
Removed qobj parent from refreshworker

sima-fastly:
remove qobj parent from refreshworker

### DIFF
--- a/src/cpp/handler/handlerengine.cpp
+++ b/src/cpp/handler/handlerengine.cpp
@@ -2070,7 +2070,7 @@ private:
 		}
 		else if(req->method() == "refresh")
 		{
-			RefreshWorker *w = new RefreshWorker(req, proxyControlClient, &cs.wsSessionsByChannel, this);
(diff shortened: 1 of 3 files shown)


src/cpp/handler/handlerengine.cpp
```diff
@@ -2070,9 +2070,9 @@ class HandlerEngine::Private : public QObject
 		}
 		else if(req->method() == "refresh")
 		{
-			RefreshWorker *w = new RefreshWorker(req, proxyControlClient, &cs.wsSessionsByChannel, this);
-			finishedConnection[w] = w->finished.connect(boost::bind(&Private::deferred_finished, this, boost::placeholders::_1, w));
-			deferreds += w;
+			auto w = std::make_unique<RefreshWorker>(req, proxyControlClient, &cs.wsSessionsByChannel);
+			finishedConnection[w.get()] = w->finished.connect(boost::bind(&Private::deferred_finished, this, boost::placeholders::_1, w.get()));
+			deferreds += w.get();
```
sima-fastly, line 2075: @jkarneges I have put this PR into draft again as I have concerns about lifetime of this connection. will address and then make this available for review then.

PR #47961, opened 2024-03-12
proxy: fix sharing with external routing

jkarneges:
Next links are fetched by the handler through the proxy, and the proxy can route these requests either internally (by looking up the target in the routes table) or externally (by making a fake route that points to anywhere). For some reason, external routing skips the inspect step, and this breaks sharing (collapsing) since inspects are necessary for sharing.

This PR fixes sharing with external routing by making sure inspects are always used.


sima-fastly (approved):
lgtm

PR #47968, opened 2024-03-26
build.rs: write conf.pri files only if different

jkarneges:
Slightly reduce build noise by not regenerating pushpin.conf.inst unless necessary.


sima-fastly (approved):
lgtm

PR #47970, opened 2024-04-25
handle chunked responses

jkarneges:
Apparently this was implemented, but we never flipped the bit to actually use it.


sima-fastly (commented):
some of the tests are getting into an infinite loop and failing

jkarneges:
Where are you seeing that? CI shows cargo audit failing.

sima-fastly (approved):

PR #47973, opened 2024-04-29
add refactored server request handler in separate module

jkarneges:
Currently, our HTTP protocol handling is split into a low level state machine (in src/http1.rs) and convenience wrappers (in src/connection.rs, alongside a bunch of other stuff). This PR refactors the server-side wrappers in src/connection.rs to support the ability to respond to invalid requests and to match the styling of the newer client-side wrappers, and places them in a new, unused module, leaving the original code alone. With the new module naming, `connection::RequestHandler` becomes `core::http1::server::Request`.

In a subsequent PR we'll update the code to use this new module and to respond to invalid requests, at which point the older code in src/connection.rs can be removed. 

The code is in a new module because src/connection.rs is getting way too big (10k+ lines), and this refactoring presented an opportunity to think about how to better organize the Rust code in the source tree. The structure I'd like to move towards is for `src` to contain one folder for each program (e.g. `src/proxy`, `src/handler`, etc) plus one folder for shared code (`src/core`). It pains me to add subfolders, but I think we need at least this minimum folder structure and we can try to keep it as flat as possible from here. This proposed structure is also very similar to our [pre-Rust folder structure](https://github.com/fastly/pushpin/tree/v1.31.0/src).

### DIFF
--- a/src/core/http1/error.rs
+++ b/src/core/http1/error.rs
@@ -1,0 +1,29 @@
+/*
+ * Copyright (C) 2024 Fastly, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+use crate::http1;
+use std::io;
+
+#[derive(Debug)]
+pub enum Error {
+    Io(io::Error),
+    Http(http1::Error),
+    RequestTooLarge(usize),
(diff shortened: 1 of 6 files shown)


src/core/http1/error.rs
```diff
@@ -0,0 +1,29 @@
...
+#[derive(Debug)]
+pub enum Error {
+    Io(io::Error),
+    Http(http1::Error),
+    RequestTooLarge(usize),
+    ResponseTooLarge(usize),
+    ResponseDuringContinue,
+    BufferExceeded,
+    Unusable,
+}
```
sima-fastly, line 29: here is a list of RFC 6455 err codes:
```
Protocol Error (1002
Unsupported Data (1003)
Abnormal Closure (1006)
Invalid Frame Payload Data (1007)
Policy Violation (1008)
Message Too Big (1009)
Mandatory Extension (1010)
Internal Server Error (1011)
TLS Handshake (1015)
```
 we might wanna consider adding these as we see fit

sima-fastly (commented):
I understand there will be a subsequent PR to this but I wonder how these updates have been tested so far

madeline-pratt (approved):
This looks good to me, especially after the explanation provided during our meeting.

sima-fastly (approved):

PR #47979, opened 2024-04-30
move client request handler into core::http1::client

jkarneges:
This moves the HTTP client code from src/connection.rs into a new module next to the new server module. With the new module naming, `connection::ClientRequest` becomes `core::http1::client::Request`. Unlike the new server module, this client module is actually used. This PR removes the code from src/connection.rs, adds the new module, and updates all references to refer to the new module. Only very minor changes were made along the way, basically enough to make it work in the new location (renaming types, marking types/fns as `pub`, wrapping field access with methods, etc).

### DIFF
--- a/src/core/http1/client.rs
+++ b/src/core/http1/client.rs
@@ -1,0 +1,604 @@
+/*
+ * Copyright (C) 2020-2023 Fanout, Inc.
+ * Copyright (C) 2023-2024 Fastly, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+use crate::buffer::{Buffer, VecRingBuffer, VECTORED_MAX};
+use crate::core::http1::error::Error;
+use crate::core::http1::util::*;
+use crate::future::{
+    select_2, AsyncRead, AsyncWrite, AsyncWriteExt, ReadHalf, Select2, StdWriteWrapper, WriteHalf,
+};
+use crate::http1;
+use crate::pin;
+use std::cell::RefCell;
+use std::io::{self, Write};
+use std::mem;
+use std::pin::Pin;
+use std::str;
+
+pub struct Request<'a, R: AsyncRead, W: AsyncWrite> {
+    r: ReadHalf<'a, R>,
+    w: WriteHalf<'a, W>,
+    buf1: &'a mut VecRingBuffer,
+    buf2: &'a mut VecRingBuffer,
+}
+
+impl<'a, R: AsyncRead, W: AsyncWrite> Request<'a, R, W> {
+    pub fn new(
+        stream: (ReadHalf<'a, R>, WriteHalf<'a, W>),
+        buf1: &'a mut VecRingBuffer,
+        buf2: &'a mut VecRingBuffer,
+    ) -> Self {
+        Self {
+            r: stream.0,
+            w: stream.1,
+            buf1,
+            buf2,
+        }
+    }
+
+    #[allow(clippy::too_many_arguments)]
+    pub fn prepare_header(
+        self,
+        method: &str,
+        uri: &str,
+        headers: &[http1::Header<'_>],
+        body_size: http1::BodySize,
+        websocket: bool,
+        initial_body: &[u8],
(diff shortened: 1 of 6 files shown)


src/core/http1/util.rs
```diff
@@ -0,0 +1,141 @@
...
+
+use crate::buffer::{Buffer, VecRingBuffer};
+use crate::future::{AsyncRead, AsyncReadExt};
+use std::cmp;
+use std::future::Future;
+use std::io;
+use std::pin::Pin;
+use std::task::{Context, Poll};
+
+pub const HEADERS_MAX: usize = 64;
```
sima-fastly, line 25: can we please leave a comment here why headers max val is 64

src/core/http1/util.rs
```diff
@@ -0,0 +1,141 @@
...
+) -> usize
+where
+    F: Fn() -> bool,
+{
+    assert!(blocks_max >= 2);
+
+    // all but one block can be used for writing
+    let allowed = blocks_max - 1;
+
+    if buf.remaining_capacity() == 0 && buf.capacity() < block_size * allowed && reserve() {
```
sima-fastly, line 42: this assumes block_size * allowed will not overflow. we can use `checked_mul` to ensure

src/core/http1/client.rs
```diff
@@ -0,0 +1,604 @@
...
+use std::io::{self, Write};
+use std::mem;
+use std::pin::Pin;
+use std::str;
+
+pub struct Request<'a, R: AsyncRead, W: AsyncWrite> {
+    r: ReadHalf<'a, R>,
+    w: WriteHalf<'a, W>,
+    buf1: &'a mut VecRingBuffer,
+    buf2: &'a mut VecRingBuffer,
```
sima-fastly, line 36: I would recommend naming these to `inputBuf` & `outputBuf` for clarity

src/core/http1/client.rs
```diff
@@ -0,0 +1,604 @@
...
+        let req = http1::ClientRequest::new();
+
+        let req_body = match req.send_header(self.buf1, method, uri, headers, body_size, websocket)
+        {
+            Ok(ret) => ret,
+            Err(_) => return Err(Error::BufferExceeded),
+        };
+
+        if self.buf2.write_all(initial_body).is_err() {
+            return Err(Error::BufferExceeded);
```
sima-fastly, line 73: we are returning Error::BufferExceeded both for failing to send headers and for failing to write the initial body to buf2. can we please differentiate

src/core/http1/client.rs
```diff
@@ -0,0 +1,604 @@
...
+    r: ReadHalf<'a, R>,
+    w: WriteHalf<'a, W>,
+    buf1: &'a mut VecRingBuffer,
+    buf2: &'a mut VecRingBuffer,
+    req_body: http1::ClientRequestBody,
+    end: bool,
+}
+
+impl<'a, R: AsyncRead, W: AsyncWrite> RequestHeader<'a, R, W> {
+    pub async fn send(mut self) -> Result<RequestBody<'a, R, W>, Error> {
```
sima-fastly, line 97: if an error occurs during the write, the buffer's read state is not rolled back, which could lead to inconsistencies. would it make sense to implement a transaction-like mechanism where buffer states can be rolled back to maintain consistency in case of errors?

src/core/http1/client.rs
```diff
@@ -0,0 +1,604 @@
...
+                |cx| {
+                    let w = &mut *inner.w.borrow_mut();
+
+                    if !w.stream.is_writable() {
+                        return None;
+                    }
+
+                    let req_body = w.req_body.take().unwrap();
+
+                    let mut buf_arr = [&b""[..]; VECTORED_MAX - 2];
```
sima-fastly, line 292: can we please document the reason for this calc

src/core/http1/client.rs
```diff
@@ -0,0 +1,604 @@
...
+}
+
+impl<'a, R: AsyncRead, W: AsyncWrite> RequestBody<'a, R, W> {
+    pub fn prepare(&self, src: &[u8], end: bool) -> Result<usize, Error> {
+        if let Some(inner) = &*self.inner.borrow() {
+            let w = &mut *inner.w.borrow_mut();
+
+            // call not allowed if the end has already been indicated
+            if w.end {
+                return Err(Error::Io(io::Error::from(io::ErrorKind::InvalidInput)));
```
sima-fastly, line 152: this errkind doesn't seem to match the case. maybe sth like `NoFurtherWritesAllowed` or 
return Err(Error::Io(io::Error::new(io::ErrorKind::Other, "No further writes allowed, end flag set")));

src/core/http1/client.rs
jkarneges: since I was only meaning to move code, let's leave this to a separate PR?

src/core/http1/client.rs
jkarneges: it's okay here since the operation can only be attempted once (`send()` consumes self)

sima-fastly (approved):
lgtm

PR #47981, opened 2024-05-03
connection: use new server module for req mode

jkarneges:
There are connection handlers in src/connection.rs for two different modes of operation: "req" mode for buffered, and "stream" mode for streaming. This PR updates the req mode handler to use the refactored server request module. The stream mode handler still uses the original server request code. The changes are being done in parts to keep the PRs smaller.

The main goal of this work is to be able to send error responses. To centralize our error handling, the req mode handler function (`server_req_handler`) is split into a few sub-functions. The most notable is `server_req_respond`, which handles a request all the way up to the point of preparing a response header. On success, a response header has been prepared for the caller to send. On error, a response header has not been prepared, allowing the caller to send an error response instead. We lose the ability to send error responses once we attempt to send a successfully prepared header, and so `server_req_respond` fully encapsulates all situations where we might want to send an error response. The code to actually send error responses when `server_req_respond` returns an error is not included and will go in a later PR.

### DIFF
--- a/src/connection.rs
+++ b/src/connection.rs
@@ -1,6 +1,6 @@
 /*
  * Copyright (C) 2020-2023 Fanout, Inc.
- * Copyright (C) 2023 Fastly, Inc.
+ * Copyright (C) 2023-2024 Fastly, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -38,8 +38,8 @@ use crate::arena;
 use crate::buffer::{
     Buffer, ContiguousBuffer, LimitBufsMut, TmpBuffer, VecRingBuffer, VECTORED_MAX,
 };
-use crate::core::http1::client;
 use crate::core::http1::Error as CoreHttpError;
+use crate::core::http1::{client, server};
 use crate::core::http1::{RecvStatus, SendStatus};
 use crate::counter::Counter;
 use crate::future::{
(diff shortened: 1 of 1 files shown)


src/connection.rs
```diff
@@ -1983,259 +1915,349 @@ where
...
     }
 
-    // determine how to respond
+    if websocket {
+        // websocket requests are not supported in req mode
 
-    let msg = {
-        let req = handler.request();
+        // toss the request body
+        body_buf.clear();
```
sima-fastly, line 1965: do we wanna add a log line here?

src/connection.rs
```diff
@@ -1983,259 +1915,349 @@ where
     }
 }
 
-// return true if persistent
+// read request body and prepare outgoing zmq message
 #[allow(clippy::too_many_arguments)]
-async fn server_req_handler<S: AsyncRead + AsyncWrite>(
+async fn server_req_read_body<R: AsyncRead, W: AsyncWrite>(
```
sima-fastly, line 1920: can we refactor this into a dedicated function to handle upgrades? to improve readability ...

sima-fastly (approved):

src/connection.rs
jkarneges: do you mean websocket upgrades? they are rejected by the req mode handler, so there is not much handling going on here

src/connection.rs
jkarneges: previously, the req mode handler would return an error response to a websocket request, which should be sufficient for visibility. I've temporarily removed that handling here in order to consolidate it with other error responses, and so it'll come back in another PR shortly after.

src/connection.rs
jkarneges: went ahead and merged. stream mode will come next, where websocket upgrades are actually accepted, and I'll see about splitting up the code there as applicable

PR #47982, opened 2024-05-07
treat some buffer writes as infallible

jkarneges:
These writes are made against memory buffers which never return an error, except for `WriteZero` which is not a real error in these cases, so we can treat the calls as infallible. Notably, this implies that `server::ResponsePrepareBody::prepare` is infallible when called with valid input.


sima-fastly (approved):
lgtm

PR #47984, opened 2024-05-07
ws err metric added to prom

sima-fastly:
adding new metrics to Prometheus 
This PR includes WS Errors


jkarneges:
Looks good as far as adding a prometheus metric goes. However what we should probably be counting here are unclean closes/disconnects rather than error signal emits, as the error signal may emit for other reasons (e.g. failure to initially connect) that we don't care about or would be better counted under different metrics.

I'm not sure, but I think all unclean closes probably emit `ErrorGeneric` which isn't too helpful since that could potentially get used in other cases. We should add a new error type, e.g. `ErrorUncleanClose`, to indicate the condition explicitly, and then only count that one in the metrics.

There should also be separate client and server metrics, since we count other metrics that way and it could be good to know.

PR #47988, opened 2024-05-14
Remove QObj from InspectRequest

sima-fastly:


sima-fastly:
test error using valGrind is showing that zrpcReq is failing to delete a deleted private d.
this is due to InspectReq inhering from ZrpcReq .
ZrpcReq we can’t remove QObject from just yet since its using QInvoke & QTimer ….
ill be closing this PR

PR #47990, opened 2024-05-15
connection: send error responses

jkarneges:
Send an error response to the client when there is a failure processing a request for any reason, rather than just closing the socket. Notably, detailed errors are sent if the request header is too big or if a WebSocket request is not well-formed.

```
% curl -i -H "Foo: `echo $(seq 1 5000) | base64`" http://localhost:7999
HTTP/1.0 400 Bad Request
Content-Type: text/plain
Content-Length: 50

Request header size exceeded limit of 8192 bytes.
```

```
% curl -i -H "Upgrade: websocket" http://localhost:7999               
HTTP/1.1 400 Bad Request
Content-Type: text/plain
Content-Length: 110

Request contained an Upgrade header with value "websocket" but the request was not a valid WebSocket request.
```

### DIFF
--- a/src/connection.rs
+++ b/src/connection.rs
@@ -1302,6 +1302,108 @@ where
     }
 }
 
+async fn send_error_response<'a, R: AsyncRead, W: AsyncWrite>(
+    mut resp: server::Response<'_, R, W>,
+    zreceiver: &TrackedAsyncLocalReceiver<'_, (arena::Rc<zhttppacket::OwnedResponse>, usize)>,
+    e: &Error,
+) -> Result<(), Error> {
+    let headers = &[http1::Header {
+        name: "Content-Type",
+        value: b"text/plain",
+    }];
+
+    let mut body: ArrayVec<u8, 512> = ArrayVec::new();
+
+    let code = match e {
+        Error::CoreHttp(CoreHttpError::Http(e)) => {
+            writeln!(&mut body, "Failed to parse request: {}", e)?;
+
+            400
+        }
+        Error::CoreHttp(CoreHttpError::RequestTooLarge(limit)) => {
+            writeln!(
+                &mut body,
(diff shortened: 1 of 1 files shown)


src/connection.rs
```diff
@@ -1302,6 +1302,108 @@ where
...
+        value: b"text/plain",
+    }];
+
+    let mut body: ArrayVec<u8, 512> = ArrayVec::new();
+
+    let code = match e {
+        Error::CoreHttp(CoreHttpError::Http(e)) => {
+            writeln!(&mut body, "Failed to parse request: {}", e)?;
+
+            400
```
sima-fastly, line 1321: can we  use `http::status::StatusCode` constants 
https://docs.rs/http/latest/http/status/struct.StatusCode.html

src/connection.rs
```diff
@@ -1606,7 +1708,12 @@ async fn server_req_handler<S: AsyncRead + AsyncWrite>(
         {
             Ok(Some(ret)) => ret,
             Ok(None) => return Ok(false), // no request
-            Err(e) => return Err(e),
+            Err(e) => {
```
sima-fastly, line 1711: what about handling err resp in other funcs such as `server_req_read_header_and_body`

src/connection.rs
jkarneges: I don't think I'd add that crate right now just for this, but it is an interesting crate we may want to consider later

src/connection.rs
jkarneges: ah, that function is called by `server_req_respond`, and so any error from it will bubble up. basically, the `*_respond` functions (`server_req_respond` and `server_stream_respond`) process a request all the way up to the moment of preparing a successful response header, such that handling their return value is the only place we need to worry about sending errors

sima-fastly (approved):

PR #47992, opened 2024-05-17
Remove qobj parent from InspectWoker

sima-fastly:
Remove qobj parent from InspectWoker

### DIFF
--- a/src/cpp/handler/handlerengine.cpp
+++ b/src/cpp/handler/handlerengine.cpp
@@ -126,16 +126,13 @@ public:
 	LastIds lastIds;
 	map<Deferred*, Connection> finishedConnection;
 
-	InspectWorker(ZrpcRequest *_req, ZrpcManager *_stateClient, bool _shareAll, QObject *parent = 0) :
-		Deferred(parent),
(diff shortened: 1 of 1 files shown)


src/cpp/handler/handlerengine.cpp
```diff
@@ -126,16 +126,13 @@ class InspectWorker : public Deferred
...
-	InspectWorker(ZrpcRequest *_req, ZrpcManager *_stateClient, bool _shareAll, QObject *parent = 0) :
-		Deferred(parent),
+	InspectWorker(ZrpcRequest *_req, ZrpcManager *_stateClient, bool _shareAll) :
 		req(_req),
 		stateClient(_stateClient),
 		shareAll(_shareAll),
 		truncated(false),
 		autoShare(false)
 	{
-		req->setParent(this);
```
jkarneges, line 137: if this is removed then need to ensure `req` is cleaned up some other way

sima-fastly: thanks for catching this. no need to remove this line here.

sima-fastly:
AcceptReq is throwing errors. need to close this for now

PR #47993, opened 2024-05-17
Remove q_obj from tests

sima-fastly:
Remove q_obj from some tests


jkarneges (commented):
Hmm, let's only remove the macro if we can also remove the inheritance.

PR #47994, opened 2024-05-17
Remove Q_Object from ServiceProcess

sima-fastly:
Remove Q_Object from ServiceProcess


jkarneges (commented):
We shouldn't remove the macro without also removing the inheritance, but here `QProcess` is not easily removed. Let's skip for now, and in the meantime maybe the new runner will take over and we won't have to worry about it.

PR #47995, opened 2024-05-17
Remove Q_Object from Mongrel

sima-fastly:
Remove Q_Object from Mongrel


jkarneges (commented):
Let's remove `QObject` from `Service` first.

PR #48002, opened 2024-05-21
ensure additional blocks are always returned

jkarneges:
Connections are only allowed to increase their buffer sizes by so many "blocks", and the available blocks are managed globally in a `Counter` (basically an atomic int with increment/decrement operations). Connections decrement the value whenever they want to use additional blocks and increment the value whenever they are done. However, in some cases the value is not being incremented when it needs to be. To fix this, and to reduce the chance of mistakes in the future, this PR adds a wrapper type `CounterDec` that keeps track of decrements and then on drop will increment by the same amount. This way, the blocks automatically get put back when the wrapper goes out of scope.

### DIFF
--- a/src/counter.rs
+++ b/src/counter.rs
@@ -133,6 +133,7 @@ mod tests {
             assert!(c.dec(1).is_err());
         }
 
-        assert!(c.dec(1).is_ok());
+        assert!(c.dec(2).is_ok());
+        assert!(c.dec(1).is_err());
     }
 }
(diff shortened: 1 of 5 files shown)


src/counter.rs
```diff
@@ -78,6 +78,30 @@ impl Counter {
     }
 }
 
+pub struct CounterDec<'a> {
+    counter: &'a Counter,
+    amount: usize,
+}
+
+impl<'a> CounterDec<'a> {
+    pub fn new(counter: &'a Counter) -> Self {
```
sima-fastly, line 87: are there any concurrency issues accessing this counter?

jkarneges: the counter is designed to be thread-safe (see the inc/dec impls). there is one instance shared by all tasks

sima-fastly (approved):

PR #48003, opened 2024-05-23
rename condure to connmgr

jkarneges:
The name "condure" (a play on "enduring connection") was a reasonable name to give when condure was an independent project, but now that it's a component of Pushpin it should be given a generic name like any other component (see: proxy, handler, etc). I believe renaming condure to something generic will help make Pushpin easier to digest.

Based on condure's role, a good generic name would be something like "connection manager" or "long-lived connection manager". In fact this is how I often refer to condure in discussion ("pushpin's connection manager can handle N connections"). But these names are a bit long for a module or executable. I can't think of any better generic names though, so I propose we use an abbreviation of connection manager: "cm". The compiled program would be named `pushpin-cm`.

The downside to "cm" is that isn't very meaningful until the reader learns what it stands for. This may seem like we're exchanging one nonsense word for another, but abbreviations are not uncommon in projects and I think once the reader learns the meaning it will be immediately comprehensible which is an improvement.

Other names I considered: connmgr, connectionmgr, connection-manager, cxnmgr, connman, conman, net, io, connio, conio.


sima-fastly (approved):
I prefer `connmgr` as a clearer alternative here.

Also in files such as `handlerapp.cpp` & `app.cpp` there are var named with a condure prefix. it would be nice to update those as well

also there are references to `condure` service in cpp runner

jkarneges:
Existing references are there for compatibility with old configs.

jkarneges:
After thinking about it, moving to connmgr :)

PR #48005, opened 2024-05-28
add StatsManager param to websocket  classes

sima-fastly:
integrate statistics management directly within the ZWebSocket & WebSocketOverHttp classes, enabling better monitoring and logging of tls handshake errors.


jkarneges:
TLS handshake errors are already exposed by the `ErrorTls` condition, so emitting metrics about this condition can happen outside of these classes, in WsProxySession's `out_error`.

PR #48006, opened 2024-05-30
when all batched items are removed, start clean

jkarneges:
`Batch` is stateful, building a table of ipc addresses when `add()` is called, and iterating over the table when `take_group()` is called. The API leaves some potential for misuse, notably calling `add()` with an address that is already in the table that has been iterated past, leading to a situation where `is_empty()` returns false yet `take_group()` returns None. Likely, this can happen in `keep_alives_task` when all items are removed during the await for channel writability and the channel turns out to not be writable and the loop continues from the start.

The `clear()` method can be used to reset everything, but having to call it after all items are removed is not intuitive and feels redundant. This PR fixes the issue by making `Batch` automatically clear its address table whenever all items have been removed. This means `clear()` is no longer required to repair the state, it simply becomes a convenience for removing all items.

A version of `Batch` exists in both client and server, so both are updated. In the future we may want to consider sharing one type.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -174,6 +174,17 @@ impl Batch {
     }
 
     fn add(&mut self, to_addr: &[u8], ckey: usize) -> Result<BatchKey, ()> {
+        if self.nodes.len() == self.nodes.capacity() {
+            return Err(());
+        }
+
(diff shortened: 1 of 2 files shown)


sima-fastly (approved):
lgtm

PR #48008, opened 2024-06-08
move connmgr and runner modules into their own dirs

jkarneges:
This begins the process of changing the source tree layout back to the style we were using pre-Rust, where each program gets its own dir. The eventual goal is for all source files to be moved into program-specific dirs, or into the core dir, such that the src dir itself does not contain any source files (except for maybe lib.rs).


sima-fastly (approved):
nice!

PR #48009, opened 2024-06-08
use std::pin::pin instead of our own macro

jkarneges:
This requires at least Rust 1.68.0.


jkarneges:
CI jobs appeared to be screwed up. Closing and making a new PR.

PR #48020, opened 2024-06-18
move various things out of lib.rs to more specific modules

jkarneges:
lib.rs ought to be kept lean, with most code living in modules. This PR is a start.


sima-fastly (approved):
lgtm

PR #48021, opened 2024-06-19
un-nest c++ files & tests, placing them next to rust files

jkarneges:
Currently, c++ files are in `src/cpp`, with tests in `src/cpp/tests`. This change flattens the source hierarchy, removing the extra dirs and placing the c++ files in the same dirs as rust files. The c++ files are placed in the core dir or program-specific dirs as appropriate.

Is this weird? Yes and no. Yes, because it's kind of odd to see rust and c++ files side by side, especially with different naming styles. Idiomatic rust names files by subject, and c++ names them by class. However, it is also not weird when considering the c++ files are essentially being moved to their original locations. Nesting the files in a subdir came about when refactoring the build system, but this nesting is no longer necessary.

Perhaps more importantly though, there's a case to be made for intentionally placing the files side by side. It helps put both languages in a more equal standing and may make maintenance easier as we increase interfacing between the two languages. Despite our increased use of rust, our use of c++ is not deprecated, and this restructure helps reflect this.


sima-fastly (approved):
big change. looks good overall

PR #48023, opened 2024-06-21
split out async channel code

jkarneges:
Similar to decentralizing our FFI code, I think core async types should be able to exist anywhere in the project instead of being centralized in a single file at the root. Putting async types in a single file maybe made sense when async was new and we wanted to limit our dependence on it, but we've been using async heavily for several years now and we're not going back so let's just embrace it.

As a first step, this PR moves AsyncSender, AsyncReceiver, etc out of `future` and into `core::channel`, alongside related non-async types.


sima-fastly (approved):
looks good if this has been tested for functionality & performance

jkarneges:
channels are covered extensively in the tests. there should be no performance impact as this is only moving where code lives

PR #48025, opened 2024-06-24
run each qt test as individual rust test

jkarneges:
This opens the door for invoking individual qt/c++ tests from anywhere in the project, rather than all of them needing to be centrally referenced in one place. It also causes `cargo test` to list them individually and allows running them individually (`cargo test proxyengine`). The tests are still regulated to run serially via a background thread, per qt requirements.

After this lands, we can look at moving these tests from the root into specific modules (e.g. move the core tests into the core dir, the proxy tests into the proxy dir, etc).


sima-fastly (approved):
glad to see this update. lgtm

PR #48026, opened 2024-06-25
move qt test calls to specific modules

jkarneges:
This moves the C++ test calls out of `lib.rs` and into the core dir or program-specific dirs as appropriate. It also moves related helper code to core. This reduces `lib.rs` to only `mod` statements and FFI imports.

```
% cargo test proxy
    Finished `test` profile [unoptimized + debuginfo] target(s) in 0.66s
     Running unittests src/lib.rs (target/debug/deps/pushpin-772dc0aebb39aa6c)

running 2 tests
test proxy::tests::proxyengine ... ok
test proxy::tests::routesfile ... ok

test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 193 filtered out; finished in 2.24s
```

Leaving the FFI imports centralized (within a `mod ffi` at the root) is probably ok, as FFI imports on the C++ side are similarly centralized (cbindgen puts them within a `namespace ffi` at the root). This should not be confused with FFI exports which based on recent code reorganization can appear anywhere.


sima-fastly (approved):
lgtm

PR #48027, opened 2024-06-25
core: use rtimer instead of qtimer for everything

jkarneges:
The max number of allowed RTimers is configured at start time, so this number needs to be adjusted in proxy & handler to cover any new usages. Currently, RTimer usage is per-session, with a limit of 10 per session (an overestimate). This PR adds an extra 100 miscellaneous timers to cover per-thread timers such as the ones in StatsManager, the default ZhttpManager, and various QZmqSockets.

For the proxy, it also adds 10 per "zroute" (a route with a zhttp target), since each zroute contains a ZhttpManager which contains timers. This required introducing a zroutes maximum. Routes with zhttp targets are likely not used much by anyone, and having a lot of them would be even more unusual, so the max is hardcoded rather than configurable.

### DIFF
--- a/src/core/qzmq/src/qzmqsocket.cpp
+++ b/src/core/qzmq/src/qzmqsocket.cpp
@@ -375,7 +375,7 @@ public:
 	bool canWrite, canRead;
 	QList< QList<QByteArray> > pendingWrites;
 	int pendingWritten;
-	RTimer *updateTimer;
+	std::unique_ptr<RTimer> updateTimer;
 	Connection updateTimerConnection;
 	bool pendingUpdate;
 	int shutdownWaitTime;
@@ -425,16 +425,13 @@ public:
 		connect(sn_read, &QSocketNotifier::activated, this, &Private::sn_read_activated);
 		sn_read->setEnabled(true);
 
-		updateTimer = new RTimer;
+		updateTimer = std::make_unique<RTimer>();
 		updateTimerConnection = updateTimer->timeout.connect(boost::bind(&Private::update_timeout, this));
 		updateTimer->setSingleShot(true);
 	}
 
 	~Private()
 	{
-		updateTimerConnection.disconnect();
-		updateTimer->deleteLater();
-
 		set_linger(sock, shutdownWaitTime);
 		wzmq_close(sock);
 
--- a/src/core/qzmq/src/qzmqsocket.cpp
+++ b/src/core/qzmq/src/qzmqsocket.cpp
@@ -28,11 +28,14 @@
 #include <assert.h>
 #include <QStringList>
 #include <QPointer>
-#include <QTimer>
 #include <QSocketNotifier>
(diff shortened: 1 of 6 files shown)


src/core/qzmq/src/qzmqsocket.cpp
```diff
@@ -421,15 +425,14 @@ class Socket::Private : public QObject
 		connect(sn_read, &QSocketNotifier::activated, this, &Private::sn_read_activated);
 		sn_read->setEnabled(true);
 
-		updateTimer = new QTimer(this);
-		connect(updateTimer, SIGNAL(timeout()), SLOT(update_timeout()));
+		updateTimer = new RTimer;
```
sima-fastly, line 428: have we considered using smart pointers for RTimer?

src/core/qzmq/src/qzmqsocket.cpp
```diff
@@ -612,6 +615,13 @@ class Socket::Private : public QObject
 		}
 	}
 
+	void update_timeout()
```
sima-fastly, line 618: this slot is still being used in the old QT style way. should we move it out of `slot` section?

src/handler/handlerengine.cpp
```diff
@@ -1327,8 +1327,8 @@ class HandlerEngine::Private : public QObject
 	{
 		config = _config;
 
-		// up to 10 timers per connection
-		RTimer::init(config.connectionsMax * 10);
+		// up to 10 timers per connection, plus an extra 100 for misc
```
sima-fastly, line 1330: why 100 ?

src/proxy/engine.cpp
```diff
@@ -380,8 +381,16 @@ class Engine::Private : public QObject
 
 	void routesChanged()
 	{
+		QList<DomainMap::ZhttpRoute> zhttpRoutes = domainMap->zhttpRoutes();
```
sima-fastly, line 384: we can always do `auto zhttpRoutes` to make the code a bit cleaner

src/core/qzmq/src/qzmqsocket.cpp
jkarneges: ah, yes I think we can do that since RTimer doesn't require `deleteLater()`.

src/core/qzmq/src/qzmqsocket.cpp
jkarneges: it is not in the slot section afaict

src/handler/handlerengine.cpp
jkarneges: just some number large enough to handle any non-session timers with potential expansion. currently there's 1 per zmq socket, and 4 per zhttpmanager. I'm guessing handler uses less than 20.

sima-fastly (approved):
lgtm

PR #48042, opened 2024-07-15
indicate when bytes are needed during server request body processing

jkarneges:
Updates the low level server logic similar to #48041. The server API style is a little different, taking a cursor as input and returning a tuple instead of an enum. In order to represent a "needs bytes" state, the tuple is wrapped in an `Option`.


sima-fastly (approved):
great improvements

PR #48043, opened 2024-07-16
call add_to_buffer when bytes needed and count recv of 0 as error

jkarneges:
Notably, this fixes a busyloop when a websocket rejection body exceeds the buffer size, returning an error instead.

For clarity, `written` is renamed to `size` when handling `try_recv()` return values. Unlike the lower level methods that return multiple byte counts, `try_recv()` only returns one count (bytes written to the output buffer), and in that case `size` is a more natural name to use.

Tested locally:

```
[DEBUG] 2024-07-16 16:30:45.604 [connmgr] [pushpin::connmgr::connection] client-conn 0-0-277d6959-6930-486d-8b76-21871ad66fb3: process error: WebSocketRejectionTooLarge(8192)
```

### DIFF
--- a/src/connmgr/connection.rs
+++ b/src/connmgr/connection.rs
@@ -1413,17 +1413,20 @@ async fn server_req_read_body<R: AsyncRead, W: AsyncWrite>(
 
     loop {
         match req_body.try_recv(body_buf.write_buf())? {
-            RecvStatus::Complete((), written) => {
-                body_buf.write_commit(written);
+            RecvStatus::Complete((), size) => {
+                body_buf.write_commit(size);
                 break;
             }
-            RecvStatus::Read((), written) => {
-                body_buf.write_commit(written);
+            RecvStatus::Read((), size) => {
+                body_buf.write_commit(size);
 
-                if written == 0 {
-                    // ABR: discard_while
-                    discard_while(zreceiver, pin!(req_body.add_to_buffer())).await?;
+                if size == 0 {
+                    return Err(Error::BufferExceeded);
                 }
+            }
+            RecvStatus::NeedBytes(()) => {
+                // ABR: discard_while
+                discard_while(zreceiver, pin!(req_body.add_to_buffer())).await?;
             }
         }
     }
@@ -2012,14 +2015,11 @@ where
                 let max_read = cmp::min(tmp_buf.len(), zsess_in.credits() as usize);
 
                 let (size, done) = match req_body.try_recv(&mut tmp_buf[..max_read])? {
-                    RecvStatus::Complete((), written) => (written, true),
-                    RecvStatus::Read((), written) => {
-                        if written == 0 {
-                            add_to_buffer.set(Some(req_body.add_to_buffer()));
-                            continue;
-                        }
-
-                        (written, false)
+                    RecvStatus::Complete((), size) => (size, true),
+                    RecvStatus::Read((), size) => (size, false),
+                    RecvStatus::NeedBytes(()) => {
+                        add_to_buffer.set(Some(req_body.add_to_buffer()));
+                        continue;
                     }
                 };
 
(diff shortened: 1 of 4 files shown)


sima-fastly (approved):
please ensure no perf hit on high load scenarios

src/connmgr/connection.rs
```diff
@@ -1413,18 +1413,21 @@ async fn server_req_read_body<R: AsyncRead, W: AsyncWrite>(
...
+                if size == 0 {
+                    return Err(Error::BufferExceeded);
                 }
             }
+            RecvStatus::NeedBytes(()) => {
+                // ABR: discard_while
+                discard_while(zreceiver, pin!(req_body.add_to_buffer())).await?;
+            }
         }
     }
```
jkarneges, line 1432: This loop reads a request body from a client into a fixed-sized buffer. Previously, it would try to read more bytes into the input buffer whenever `try_recv` returned zero, which could happen whenever bytes were needed or if the output buffer was full. Now, it only tries to read more bytes into the input buffer whenever `try_recv` returns `NeedBytes`.

This means the only change in behavior is what it does when the output buffer is full. Previously, it would try to read more bytes into the input buffer, either hanging the connection or eventually returning `BufferExceeded` if the input buffer filled up also. Now, it immediately returns `BufferExceeded`. So there should be no performance impact during regular operation, just an earlier error when the output is full and avoiding a connection hang.

src/connmgr/connection.rs
```diff
@@ -2012,14 +2015,11 @@ where
...
-                        }
-
-                        (written, false)
+                    RecvStatus::Complete((), size) => (size, true),
+                    RecvStatus::Read((), size) => (size, false),
+                    RecvStatus::NeedBytes(()) => {
+                        add_to_buffer.set(Some(req_body.add_to_buffer()));
+                        continue;
                     }
                 };
```
jkarneges, line 2024: This code reads request body bytes from a client into a temporary buffer that is guaranteed to be non-zero-sized. Previously, it would try to read more bytes into the input buffer whenever `try_recv` returned zero, which could only happen whenever bytes were needed since the output buffer would always have room. Now, it only tries to read more bytes into the input buffer whenever `try_recv` returns `NeedBytes`. This is identical behavior as before so there should be no performance impact.

src/connmgr/connection.rs
```diff
@@ -2110,14 +2110,11 @@ where
...
-                        }
-
-                        (written, None)
+                    RecvStatus::Complete(finished, size) => (size, Some(finished)),
+                    RecvStatus::Read((), size) => (size, None),
+                    RecvStatus::NeedBytes(()) => {
+                        add_to_buffer.set(Some(resp_body.add_to_buffer()));
+                        continue;
                     }
                 };
```
jkarneges, line 2119: This is similar to the code that reads request body bytes from a client into a temporary buffer that is guaranteed to be non-zero-sized, except this time it's for reading response body bytes from a server. The updates have identical behavior as before so there should be no performance impact.

src/connmgr/connection.rs
```diff
@@ -4583,18 +4580,19 @@ where
...
 
-                        if written == 0 {
-                            resp_body.add_to_buffer().await?;
+                        if size == 0 {
+                            return Err(Error::BufferExceeded);
                         }
                     }
+                    RecvStatus::NeedBytes(()) => resp_body.add_to_buffer().await?,
                 }
             }
```
jkarneges, line 4597: This is similar to the code that reads request body bytes from a client into a fixed-size buffer, except this time it's for reading response body bytes from a server.

This means the only change in behavior is what it does when the output buffer is full. Previously, it would try to read more bytes into the input buffer, either hanging the connection, hanging the thread (busylooping), or eventually returning `BufferExceeded` if the input buffer filled up also. Now, it immediately returns `BufferExceeded`. So there should be no performance impact during regular operation, just an earlier error when the output is full and avoiding hangs.

src/connmgr/connection.rs
```diff
@@ -5185,30 +5183,29 @@ where
...
+                                        Select2::R1(ret) => {
+                                            ret?;
+                                            break;
+                                        }
+                                        Select2::R2(ret) => {
+                                            let zreq = ret?;
+
+                                            // ABR: handle_other
+                                            server_handle_other(zreq, zsess_in, zsess_out).await?;
                                         }
```
jkarneges, line 5209: This is similar to earlier code that reads response body bytes from a server into a buffer that is guaranteed to be non-zero-sized. The updates have identical behavior as before so there should be no performance impact.

src/connmgr/connection.rs
```diff
@@ -5298,41 +5295,36 @@ where
...
                                     }
                                 }
```
jkarneges, line 5330: This is similar to earlier code that reads response body bytes from a server into fixed-sized buffer.

This means the only change in behavior is what it does when the output buffer is full. Previously, it would try to read more bytes into the input buffer, either hanging the connection, hanging the thread (busylooping), or eventually returning `WebSocketRejectionTooLarge` if the input buffer filled up also. Now, it immediately returns `WebSocketRejectionTooLarge`. So there should be no performance impact during regular operation, just an earlier error when the output is full and avoiding hangs.

jkarneges:
> please ensure no perf hit on high load scenarios

Double-checked the code and left some comments.

PR #48054, opened 2024-08-03
Update time crate

jannic:
time 0.3.18 is incompatible with rust 1.80.0. This has been fixed in 0.3.36.

This fixes the following error message:
```
   Compiling time v0.3.18
error[E0282]: type annotations needed for `Box<_>`
  --> /home/jan/.cargo/registry/src/index.crates.io-6f17d22bba15001f/time-0.3.18/src/format_description/parse/mod.rs:83:9
   |
83 |     let items = format_items
   |         ^^^^^
...
86 |     Ok(items.into())
   |              ---- type must be known at this point
   |
help: consider giving `items` an explicit type, where the placeholders `_` are specified
   |
83 |     let items: Box<_> = format_items
   |              ++++++++

For more information about this error, try `rustc --explain E0282`.
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,7 +41,7 @@ signal-hook = "0.3"
 slab = "0.4"
 socket2 = "0.4"
 thiserror = "1.0"
-time = { version = "=0.3.18", features = ["formatting", "local-offset", "macros"] }
+time = { version = "0.3.36", features = ["formatting", "local-offset", "macros"] }
 url = "2.3"
 zmq = "0.9"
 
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -41,7 +41,7 @@ signal-hook = "0.3"
 slab = "0.4"
 socket2 = "0.4"
 thiserror = "1.0"
-time = { version = "=0.3.18", features = ["formatting", "local-offset", "macros"] }
```
sima-fastly, line 44: thanks for the udpate.
lets pin to the specific version `version "=0.3.36"`

Cargo.toml
```diff
@@ -52,7 +52,7 @@ test-log = "0.2"
 
 [build-dependencies]
 pkg-config = "0.3"
-time = { version = "=0.3.18", features = ["formatting", "local-offset", "macros"] }
```
sima-fastly, line 55: same here

Cargo.toml
jannic: What's the advantage of pinning to a specific version? Most of the dependencies don't do it, so I guess the exceptions are there for a reason, but I don't recognize that reason for the `time` crate.

jkarneges (approved):

PR #48056, opened 2024-08-12
Rework CI

deg4uss3r:
This PR does multiple things but has been tested over in my fork: https://github.com/deg4uss3r/pushpin/pull/1

## Changes 

1. Moving the dependencies to a docker image, probably the most controversial change
2. Moving CodeQL to an "Advanced" version and as an action that can run on each PR (and also fixed the c++ analysis)
3. Added the Test job which did not exist prior
4. Updated to running on the stable version of Rust (due to Homebrew restrictions it always builds with stable so I believe we should as well)
5. Bumped time to `0.3.36` to comply with Rust `1.80.0` 
6. Added the Rust cfg flag for `qt_lib_prefix` for Rust `1.80.0` 

## Timings

A simple analysis of the timings has found this to be faster both on average for a successful run as well as a better way to bail out quickly on failures.

### Success Runs

Overall it looks to be on average faster the average time on the `main` branch varies greatly from ~28 minutes to ~16 minutes, it appears to be more common to be in the high 20 time frame.

https://github.com/fastly/pushpin/actions/workflows/test.yml?query=is%3Asuccess

The changes here (over 3 runs so the sample is small) is tightly within the 17 minute mark. This also includes a test run which did not exist previously. 

https://github.com/deg4uss3r/pushpin/actions/workflows/suite.yml?query=is%3Asuccess

### Failure Runs 

Gains here are the quick failures due to the dependency flow on jobs now. We won't spend extra time running additional jobs if the check, lint, etc fail before more expensive jobs run. 

The current flow: https://github.com/fastly/pushpin/actions/workflows/test.yml?query=is%3Afailure 
The new flow: https://github.com/deg4uss3r/pushpin/actions/workflows/suite.yml?query=is%3Afailure


Cargo.toml
```diff
@@ -41,7 +41,7 @@ signal-hook = "0.3"
 slab = "0.4"
 socket2 = "0.4"
 thiserror = "1.0"
-time = { version = "0.3.36", features = ["formatting", "local-offset", "macros"] }
+time = { version = "=0.3.36", features = ["formatting", "local-offset", "macros"] }
```
jkarneges, line 44: Is there a reason to pin the full version?

Cargo.toml
```diff
@@ -97,3 +97,7 @@ bench = false
 name = "pushpin-publish"
 test = false
 bench = false
+
+[lints.rust]
+
+unexpected_cfgs = { level = "warn", check-cfg = ['cfg(qt_lib_prefix, values("Qt", "Qt6", "Qt5"))'] }
```
jkarneges, line 103: Got a separate PR out for this: https://github.com/fastly/pushpin/pull/48059

Cargo.toml
deg4uss3r: I usually pin to the exact version if the crate is under `1.0.0` to avoid breaking changes

.github/Dockerfile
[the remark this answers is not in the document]
deg4uss3r: Correct, since it's used only in CI I left the file here, it can go where ever we want, I could also create a `fanout-team` Docker user or something like that and git-crypt or share the credentials in a 1Password vault?

Cargo.toml
deg4uss3r: Fixed to use `main`'s version.

.github/Dockerfile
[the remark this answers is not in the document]
deg4uss3r: Fixed to use `fanout/build-base` image rather than one I owned.

deg4uss3r:
This PR is now ready for full review, I updated the `Makefile` to include a way to select which rust toolchain is ran, that is the major logic change since the original PR was opened as well as updated the caching to work better so we'll download less crates for all of the jobs. 

Finally, we are now also running our minimum supported rust version (`1.70.0`) in parallel with the latest stable so we will now if we break in stable or a change breaks our minimum supported build.

deg4uss3r:
Okay this is now ready for final review, the most recent commits: 

- fixed the `Makefile` to use the default environment Rust (e.g. does not force stable) and only prints if you explicitly set it 
- The `Makefile` also uses `TOOLCHAIN` in the `cargo-test` stage 
- `Makefile` `cargo-test` has the `--all-features` flag which does the same it does now but future-proofs testing against all features we might add or use
- `CodeQL` now uses the official `Fanout/build-base` docker image rather than the old personalized one
- The CI jobs were updated to use `make` when there is a `Make` command that does the same as it was running
- The CI/CodeQL suite utilized a new caching action that better and more succinctly does what we were doing previously

jkarneges (approved):

PR #48060, opened 2024-08-19
fix yield_to_local_events() by ignoring wakes until ready

jkarneges:


deg4uss3r (approved):
Adding my approval-lite here the code and flow look good to me.

PR #48065, opened 2024-09-05
httpsession: prep work for supporting fetches other than next

jkarneges:
This reworks the next link handling in `HttpSession` to make it reusable for other things in the future. It should not introduce any change in behavior. Some `if(state == Proxying)` conditionals are added in places where `state` is known/assumed to already be of that value.

### DIFF
--- a/src/handler/httpsession.cpp
+++ b/src/handler/httpsession.cpp
@@ -154,7 +154,7 @@ public:
 	RTimer *retryTimer;
 	StatsManager *stats;
 	ZhttpManager *outZhttp;
-	ZhttpRequest *outReq; // for fetching next links
+	std::unique_ptr<ZhttpRequest> outReq; // for fetching links
 	RateLimiter *updateLimiter;
 	PublishLastIds *publishLastIds;
 	HttpSessionUpdateManager *updateManager;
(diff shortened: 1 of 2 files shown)


src/handler/httpsession.cpp
```diff
@@ -1232,114 +1242,121 @@ class HttpSession::Private : public QObject
 				return;
 			}
 
-			if(outReq->bytesAvailable() > 0)
+			if(state == Proxying)
```
jkarneges, line 1245: this section is simply an indentation (and assertion of the state value)

src/handler/httpsession.cpp
```diff
@@ -1477,21 +1494,24 @@ private slots:
 		{
 			haveOutReqHeaders = true;
 
-			// apply ProxyContent filters of all channels
-			QStringList allFilters;
-			foreach(const Instruct::Channel &c, instruct.channels)
+			if(state == Proxying)
```
jkarneges, line 1497: another indentation

src/handler/httpsession.cpp
```diff
@@ -1501,34 +1521,38 @@ private slots:
...
 
-		delete responseFilters;
-		responseFilters = 0;
-
-		delete outReq;
-		outReq = 0;
+		cleanupOutReq();
 
-		log_debug("httpsession: failed to retrieve next link");
+		if(state == Proxying)
```
jkarneges, line 1526: another indentation based on state value

deg4uss3r (approved):
The moves in here look good to me, and pretty easily traceable, thanks Justin!

PR #48078, opened 2024-10-02
connmgr: add ability to respond via router instead of pub

jkarneges:
Related to #48073, this completes the implementation in connmgr for sending response data via router. After this, proxy will need to be updated to support receiving response data this way.

Notably, the following changes are made:

1. `make_zhttp_response()` is reworked to support producing either PUB-style or ROUTER-style message formats, returning a `Some(addr)` in the case of the latter. It is also made public and shared across modules to reduce code duplication.
2. If router-resp is indicated on the request, a flag is set to true on the `StreamSharedData` struct. Whenever response data needs to be sent, this flag can be checked to decide whether to respond via ROUTER or not.
3. When adding to a batch, specify whether router mode should be used.
4. Instead of always specifying `None` for the address, the value is specified as `Some` or `None` depending on whether router-resp was requested.

The rest of the changes are mainly tests.

### DIFF
--- a/src/connmgr/client.rs
+++ b/src/connmgr/client.rs
@@ -17,7 +17,8 @@
 
 use crate::connmgr::batch::{Batch, BatchKey};
 use crate::connmgr::connection::{
-    client_req_connection, client_stream_connection, ConnectionPool, StreamSharedData,
+    client_req_connection, client_stream_connection, make_zhttp_response, ConnectionPool,
+    StreamSharedData,
 };
 use crate::connmgr::counter::Counter;
 use crate::connmgr::resolver::Resolver;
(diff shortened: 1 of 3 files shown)


src/connmgr/client.rs
```diff
@@ -420,46 +425,30 @@ impl Connections {
 
             assert!(count <= zhttppacket::IDS_MAX);
 
-            let zreq = zhttppacket::Request {
+            let zresp = zhttppacket::Response {
```
jkarneges, line 428: This should be a response packet here rather than a request packet. We've simply been lucky that they serialize the same.

madeline-pratt (approved):

PR #48088, opened 2024-12-12
handler: make message filtering async-capable

jkarneges:
This introduces the `MessageFilter` class for implementing async filtering operations, specifically for messages. Response content (formerly "proxy" content) must still be filtered using the synchronous `Filter` API.

There are no actual async filters implemented yet, although all of the existing synchronous filters have been updated to derive from `MessageFilter` and emit async results immediately. This way we can start using the async API in connection handlers.

Longer term, the plan is to introduce a `ResponseFilter` class for similarly handling response content, and then turn `Filter` into a namespace. In preparation for this, `MessageFilter` and related types are nested under the `Filter` class.

The goal of this reworking is not only to enable async operations but also to make the filtering API more clear. Instead of a single class with different modes, we'll have different classes with single modes.

### DIFF
--- a/src/cpp.pro
+++ b/src/cpp.pro
@@ -1,6 +1,6 @@
 TEMPLATE = lib
 CONFIG -= app_bundle
-CONFIG += staticlib c++14
+CONFIG += staticlib c++17
 QT -= gui
 QT += network
 TARGET = pushpin-cpp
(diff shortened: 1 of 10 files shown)


deg4uss3r (approved):
Looks good to me, thanks for including the test!

PR #48089, opened 2024-12-18
add http-check and http-modify filters

jkarneges:
This adds two new message filters that allow delegating filtering logic to an HTTP server:

* `http-check`: Sends metadata only (no message content) to the server, to determine the action (send or drop).
* `http-modify`: Sends metadata and message content to the server, to determine the action and potentially have the content modified.

For both filters, the target endpoint is specified using the subscription metadata field `url`. Requests use the `POST` method and contain `Sub-Meta`, `Pub-Meta`, and `Grip-Last` headers. For the `http-modify` filter, the message content is included in the body (`http-check` filter leaves the body empty). The server is expected to reply with a status code 200 or 204, and optionally an `Action: drop` header to indicate a drop action. For `http-modify`, code 200 means the replacement content is in the response body, and code 204 means no change to the content.

### DIFF
--- a/src/handler/filter.cpp
+++ b/src/handler/filter.cpp
@@ -23,9 +23,13 @@
 
 #include "filter.h"
 
+#include <QJsonDocument>
+#include <QJsonObject>
 #include "log.h"
 #include "format.h"
 #include "idformat.h"
+#include "zhttpmanager.h"
+#include "zhttprequest.h"
 
 namespace {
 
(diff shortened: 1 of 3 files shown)


deg4uss3r (approved):

src/handler/filter.cpp
```diff
@@ -304,6 +308,202 @@ class VarSubstFilter : public Filter, public Filter::MessageFilter
...
+		HttpHeaders headers;
+
+		{
+			QVariantMap vmap;
+			QHashIterator<QString, QString> it(context.subscriptionMeta);
+			while(it.hasNext())
+			{
+				it.next();
+				vmap[it.key()] = it.value();
+			}
```
mhp, line 394: This feels like an odd construction.  I don't have the context on exactly what you are doing here but it looks like you are skipping the first item in the iterator by calling `it.next()` before copying.  Is that intentional?  I guess it must be, but maybe a comment as to what you are skipping might help make this look a little less odd!  Same with the two loops that follow.

mhp (commented):
I don't have the context to fully understand what's going on here, but there was one bit that stood out as potentially odd.  Hope that's helpful!

src/handler/filter.cpp
jkarneges: The iterator requires calling `next()` to access the first item, so the code is intentional and not skipping the first item. It's a quirk of Qt's unintuitive "Java-style" API:

> Unlike STL-style iterators, Java-style iterators point between items rather than directly at items. The first call to [next](https://doc.qt.io/qt-6/qhashiterator.html#next)() advances the iterator to the position between the first and second item, and returns the first item; the second call to [next](https://doc.qt.io/qt-6/qhashiterator.html#next)() advances the iterator to the position between the second and third item; and so on.

mhp: Huh!  What an odd interface!  Well, that makes sense - thanks for explaining it.

PR #48091, opened 2025-01-07
format python programs with black

jkarneges:
```
black src/m2adapter/tools tools
```


deg4uss3r:
We should probably include a check within the CI suite: 

https://fastly.slack.com/archives/C8YDS38KY/p1736272716974899?thread_ts=1736270245.550339&cid=C8YDS38KY

deg4uss3r (approved):

PR #48101, opened 2025-01-22
add DeferCall to replace QMetaObject::invokeMethod

jkarneges:
This helps reduce direct dependence on `QObject` by wrapping `QMetaObject::invokeMethod` with our own internal API for doing the same. It introduces the `DeferCall` type and updates a few places to use it. The goal is to use `DeferCall` for all delayed invocations in the entire project, such that the only direct usage of `invokeMethod` occurs within `DeferCall` itself. Then when we replace the Qt event loop we can substitute out that final usage.

Unlike `invokeMethod`, which takes a method name by string and a variable number of dynamically typed args, `DeferCall::defer` simply takes a function/closure. This allows static type-checking, and any data that would have normally been supplied by args can be captured instead.

A single `DeferCall` instance can be used to queue multiple calls. Destroying it will cancel the calls. Objects that need to make deferred calls to themselves are expected to do it through a `DeferCall` kept as a member variable. That way, when the object is destroyed so are its queued calls.

`DeferCall` also provides a replacement for `QObject::deleteLater`, via the static method `DeferCall::deleteLater`. Deferred deletions are often applied to child objects during parent object destruction, so it is important for the queued calls to remain in effect after the parent object goes away. To help with that, these calls are made against a singleton (thread local) instead of a member instance. `DeferCall::cleanup` is used to destroy the singleton at thread/program exit.

Tested with valgrind and came up clean.

### DIFF
--- a/src/core/defercall.cpp
+++ b/src/core/defercall.cpp
@@ -1,0 +1,65 @@
+/*
+* Copyright (C) 2025 Fastly, Inc.
+*
+* This file is part of Pushpin.
+*
+* $FANOUT_BEGIN_LICENSE:APACHE2$
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*
+* $FANOUT_END_LICENSE$
+*/
+
+#include "defercall.h"
+#include <assert.h>
+
+static thread_local DeferCall *g_instance = nullptr;
+
+DeferCall::DeferCall() = default;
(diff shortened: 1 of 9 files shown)


deg4uss3r (approved):

src/core/defercall.cpp
```diff
@@ -0,0 +1,65 @@
...
+	g_instance = nullptr;
+}
+
+void DeferCall::callNext()
+{
+	// there can't be more invokeMethod resolutions than queued calls
+	assert(!deferredCalls_.empty());
+
+	Call c = deferredCalls_.front();
+	deferredCalls_.pop_front();
```
mhp, line 62: This will execute a list of deferred calls in First-In-First-Out order, which seems to make sense to me.  But `defer` in go executes in [reverse order](https://go.dev/ref/spec#Defer_statements).  I can't recall the specifics of any other languages right now, but I wonder if there's an expectation for one or other that might cause problems in the future, especially when using the `deleteLater` feature with complex data structures.

jkarneges: Confirmed the deletions are in-order.

There is one exception: if there are nested event loops, deferred deletions only occur within the event loop that they were originally queued. So processing of deletions within an event loop is FIFO, but processing of the loops themselves is LIFO. In any case, we don't use nested event loops.

PR #48111, opened 2025-02-04
Bump openssl to 0.10.70 to handle RUSTSEC-2025-0004

jkarneges:
Used rust 1.82 to generate Cargo.lock to keep it on version 3 for compatibility with rust 1.75.


deg4uss3r (approved):

deg4uss3r:
Approval still stands, the difference only being the lock file version update with newer versions of Rust, this diff is better than the previous.

PR #48120, opened 2025-02-14
Use Single Thread Pool For Testing

deg4uss3r:
I was able to reproduce some hangs during testing but this seems to have worked for me locally. Trying in CI.


deg4uss3r:
Bah, looks like we are consistently still locking up the runner. I wonder if this has to do with something in CI?

PR #48132, opened 2025-02-20
defercall: use timers instead of QMetaObject::invokeMethod

jkarneges:
This reworks `DeferCall` to trigger calls using timers of zero rather than relying on a special mechanism provided by the event loop. Notably, this makes `DeferCall` work with the new event loop, since timers work with the new event loop.

A side effect is the timer subsystem must be initialized before queuing any deferred calls, and all deferred calls must be resolved/cleared before the timer subsystem is deinitialized.


mhp (approved):
LGTM!

PR #48133, opened 2025-02-22
eventloop: various registration management fixes

jkarneges:
This fixes a few things:

* Reactor registrations are set up to use a persistent waker. Without this, the reactor triggers the waker only once and then discards it.
* Registration readiness is cleared after processing. Without this, the reactor won't trigger the registration's waker again after the first event.
* Timer registrations are automatically removed after activation.

### DIFF
--- a/src/core/eventloop.rs
+++ b/src/core/eventloop.rs
@@ -62,7 +62,7 @@ impl Evented {
 }
 
 struct Registration<C> {
-    _evented: Evented,
+    evented: Evented,
     activated: bool,
     callback: Option<C>,
 }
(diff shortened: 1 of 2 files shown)


src/core/eventloop.rs
```diff
@@ -338,8 +360,11 @@ impl<C: Callback> EventLoop<C> {
             .expect("slab should have capacity"))
     }
 
-    pub fn deregister(&self, id: usize) {
-        self.regs.remove(id);
+    pub fn deregister(&self, id: usize) -> Result<(), EventLoopError> {
+        match self.regs.remove(id) {
+            Ok(()) => Ok(()),
+            Err(_) => Err(EventLoopError),
+        }
```
deg4uss3r, line 367: This can be simplified down since we're just throwing away the error:

```suggestion
 self.regs.remove(id).map_err(|_| EventLoopError)
```

deg4uss3r (approved):
Just 1 nit.

PR #48136, opened 2025-02-25
add ffi for tcpstream

jkarneges:
To move off of the Qt event loop, we'll need a replacement for `QTcpServer`/`QTcpSocket`. In the spirit of building on top of Rust, let's wrap `std::net::{TcpListener, TcpStream}` in an FFI, as opposed to using some other C++ TCP lib or the C socket API directly.


deg4uss3r (approved):
Look good to me!

PR #48141, opened 2025-02-28
finish c++ wrapper for tcpstream

jkarneges:
Following up on #48140, this finishes the `TcpStream` class.

Instead of designing it like `QTcpSocket`, which has internal buffering and special state change signals, I opted to make it a more direct wrapper around the underlying FFI / Rust type. The `read` and `write` methods perform immediate non-blocking I/O and return `errno` codes (with reads of zero indicating end-of-stream), and the signals act like edge-triggered I/O notifications (even if the underlying notifiers act level-triggered, as `QSocketNotifier` does under the Qt event loop). I think this will be a better place to evolve from as we try to converge Rust & C++ behavior, even if it means the API isn't a drop-in substitute.

### DIFF
--- a/src/core/tcplistener.cpp
+++ b/src/core/tcplistener.cpp
@@ -69,8 +69,7 @@ std::unique_ptr<TcpStream> TcpListener::accept()
 	if(!s_inner)
 		return std::unique_ptr<TcpStream>(); // null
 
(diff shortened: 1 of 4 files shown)


mhp (approved):
Looks like a pragmatic interface to me!

PR #48145, opened 2025-03-04
socketnotifier: support read & write interest at the same time

jkarneges:
We originally modeled `SocketNotifier` after `QSocketNotifier`, which takes only one kind of I/O interest at a time. If both read and write interest are needed, then two notifiers must be created. However, creating two notifiers (and thus, two registrations) for the same file descriptor can cause problems with our new event loop, which is based on `epoll` (via the `mio` crate). This PR addresses this issue by updating `SocketNotifier` to support multiple interests in a single instance, avoiding the need for multiple registrations of the same file descriptor when the new event loop is used.

Notably, this change means the `activated` signal needs to indicate which interest caused the callback (which we call "readiness" in that context). Also, when the Qt event loop is used and multiple interests are indicated then we need to use multiple `QSocketNotifier` instances as backing.

### DIFF
--- a/src/core/socketnotifier.h
+++ b/src/core/socketnotifier.h
@@ -36,8 +36,8 @@ public:
 	SocketNotifier(int socket, uint8_t interest);
 	~SocketNotifier();
 
-	bool isReadEnabled() const { return writeEnabled_; }
-	bool isWriteEnabled() const { return readEnabled_; }
+	bool isReadEnabled() const { return readEnabled_; }
+	bool isWriteEnabled() const { return writeEnabled_; }
 	int socket() const { return socket_; }
 
 	void setReadEnabled(bool enable);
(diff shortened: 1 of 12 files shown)


src/core/socketnotifier.h
```diff
@@ -27,36 +27,40 @@ class SocketNotifier : public QObject
...
+		Write = 0x02,
 	};
 
-	SocketNotifier(int socket, Type type);
+	SocketNotifier(int socket, uint8_t interest);
 	~SocketNotifier();
 
-	bool isEnabled() const { return enabled_; }
+	bool isReadEnabled() const { return writeEnabled_; }
+	bool isWriteEnabled() const { return readEnabled_; }
```
deg4uss3r, line 40: Are these flipped?

deg4uss3r (changes requested):

mhp (approved):

deg4uss3r (approved):

PR #48147, opened 2025-03-05
tcpstream: test against both event loops

jkarneges:
This updates the tests to test against both the Qt event loop and the new event loop. In order to do this, the ability to make outbound connections is added to `TcpStream`, and then `TcpStream` is used as the connecting client instead of `QTcpSocket` since it works with both loops. The tests are also updated to properly treat the API as edge-triggered, always assuming I/O operations are possible until told otherwise by `EAGAIN`.

### DIFF
--- a/src/core/net.rs
+++ b/src/core/net.rs
@@ -754,7 +754,7 @@ mod ffi {
     use std::convert::TryInto;
     use std::ffi::{CStr, OsStr};
     use std::io::{Read, Write};
-    use std::os::fd::AsRawFd;
+    use std::os::fd::{AsRawFd, FromRawFd, IntoRawFd};
     use std::os::raw::{c_char, c_int};
     use std::os::unix::ffi::OsStrExt;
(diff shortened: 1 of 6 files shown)


src/core/net.rs
```diff
@@ -945,6 +990,33 @@ mod ffi {
         }
     }
 
+    #[allow(clippy::missing_safety_doc)]
+    #[no_mangle]
+    pub unsafe extern "C" fn tcp_stream_check_connected(
+        s: *const TcpStream,
+        out_errno: *mut c_int,
+    ) -> c_int {
+        let s = s.as_ref().unwrap();
```
mhp, line 999: Your earlier fn checked `out_errno` - is it worth adding the same check to this one?

```suggestion
        assert!(!out_errno.is_null());
        
        let s = s.as_ref().unwrap();
```

mhp (approved):

PR #48153, opened 2025-03-10
fix flaky test

jkarneges:
When monitoring a zmq socket for I/O events, the socket's event state must be re-retrieved after sending a message. This test was forgetting to do that, causing it to miss events sometimes and hang.


deg4uss3r (approved):
I can't believe it was that!

PR #48165, opened 2025-03-18
proxy/handler: don't use QObject parenting with anything from core

jkarneges:


src/proxy/zroutes.cpp
```diff
@@ -120,7 +120,7 @@ class ZRoutes::Private : public QObject
 	{
 		if(!defaultItem)
 		{
-			ZhttpManager *manager = new ZhttpManager(this);
+			ZhttpManager *manager = new ZhttpManager;
```
jkarneges, line 123: cleanup of these instances are handled manually

deg4uss3r (approved):

PR #48174, opened 2025-03-21
deferred: don't delete self, to allow smart pointers to manage

jkarneges:
The `Deferred` class provides a generic interface for an async operation, signaling a `QVariant` result. Currently it deletes itself upon completion (literally with `delete this`), intended to enable a fire-and-forget developer experience when used along with `QObject` parenting. The idea is whether it completes or its parent goes away, it is automatically cleaned up, without the caller needing to explicitly track it. However, we are trying to avoid `QObject` parenting, preferring to use smart pointers instead, and self-destruction is incompatible with smart pointers.

In order to allow `Deferred` instances to be managed with smart pointers, this PR removes the self destruction and updates all the places `Deferred` is used to ensure instances are destroyed on completion or when the parent/owner goes away. In some places we were already using smart pointers to manage some instances, which was almost certainly wrong (possibly leading to double-deletions) and this PR fixes those cases too.

The resulting developer experience is not as nice but at least this fixes some wrongness and helps get us off `QObject`.

### DIFF
--- a/src/handler/refreshworker.h
+++ b/src/handler/refreshworker.h
@@ -27,18 +27,20 @@
 #include <QByteArray>
 #include <QHash>
 #include <QSet>
-#include "deferred.h"
 #include <boost/signals2.hpp>
+#include "deferred.h"
+#include "zrpcrequest.h"
 
 using Connection = boost::signals2::scoped_connection;
 
-class ZrpcRequest;
 class ZrpcManager;
 class StatsManager;
 class WsSession;
 
 class RefreshWorker : public Deferred
 {
+	Q_OBJECT
+
 public:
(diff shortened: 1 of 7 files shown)


src/handler/refreshworker.h
```diff
@@ -27,18 +27,20 @@
...
 using Connection = boost::signals2::scoped_connection;
 
-class ZrpcRequest;
 class ZrpcManager;
 class StatsManager;
 class WsSession;
 
 class RefreshWorker : public Deferred
 {
+	Q_OBJECT
```
jkarneges, line 42: Technically this should be here since `Deferred` inherits from `QObject`, but probably not for much longer.

deg4uss3r (approved):

PR #48177, opened 2025-03-21
handlerengine: don't use QObject parenting

jkarneges:


src/handler/handlerengine.cpp
```diff
@@ -841,10 +839,6 @@ class AcceptWorker : public Deferred
 		// swap instead of std::move since sessions is a member and should have a known state
 		QList<std::shared_ptr<HttpSession>> out;
 		out.swap(sessions);
-
-		foreach(const std::shared_ptr<HttpSession> &hs, out)
-			hs->setParent(0);
```
jkarneges, line 846: These objects are known to be created with no parent to begin with.

mhp (approved):

PR #48184, opened 2025-03-27
proxy: don't use QObject parenting

jkarneges:


src/proxy/proxysession.cpp
```diff
@@ -234,7 +234,6 @@ class ProxySession::Private : public QObject
 
 		SessionItem *si = new SessionItem;
 		si->rs = rs;
-		si->rs->setParent(this);
```
jkarneges, line 237: The `rs` field is manually cleaned up in `ProxySession::Private::~Private`.

mhp (approved):

PR #48185, opened 2025-03-28
websocketoverhttp: support keeping back data until more arrives

jkarneges:
Implements #48187. Basically, instead of calling `removeContentFromFrames()` before sending the exported events to the backend, it is called after receiving the response from the backend, and instead of removing all exported content only the amount accepted is removed.

Notes:

* It is not possible to keep back a `CLOSE` message, since there won't be any data after it to wait for. If Pushpin sends a request containing a `CLOSE`, the backend must accept all of the content in the request or the session will be terminated.
* The amount of data that can be kept back is limited to the size of Pushpin's buffer. If exceeded (as determined by the inability to make progress), the session will be terminated.
* The buffer is always replayed in the next request, even if the next request is not caused by the arrival of more data from the client. Using `Keep-Alive-Interval` or publishing a `refresh` action will cause the buffer to be replayed.
* When Pushpin replays data, it includes a `Content-Bytes-Replayed` header in the request. The backend can use this to confirm whether the use of `Content-Bytes-Accepted` worked.

### DIFF
--- a/src/proxy/websocketoverhttp.cpp
+++ b/src/proxy/websocketoverhttp.cpp
@@ -207,6 +207,7 @@ public:
 	int reqPendingBytes;
 	int reqFrames;
 	int reqContentSize;
+	bool reqMaxed;
 	bool reqClose;
 	int reqCloseContentSize;
 	BufferList inBuf;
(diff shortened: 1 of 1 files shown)


mhp (approved):
I checked the various bits of bookkeeping and it looked right to me.  How easy is it to add a couple of test cases, to make sure this continues to work?

jkarneges:
Tests for these changes won't be easy since they'll need to be async against a fake backend and we don't have any such tests like that for this class yet, but this is a good occasion to remedy that. Thanks for approving. I'll land and figure out how to add tests.

PR #48191, opened 2025-04-01
websocketoverhttp: fix written/writable indicators and cleanup hang

jkarneges:
This fixes a few issues:

* `writeBytesAvailable()` now doesn't count the size of an active request against the buffered amount, since we leave the bytes in the buffer until the backend accepts.
* `framesWritten()` now emits only the amount of content accepted by the backend, rather than the entire content of a request. Notably, if partial content of a frame is accepted, it will emit a count of 0 frames (and a >0 count of content bytes). Currently nothing is using this signal so these changes don't break anything.
* `clearDisconnectManager()` no longer hangs when there are managed objects left.

### DIFF
--- a/src/proxy/websocketoverhttp.cpp
+++ b/src/proxy/websocketoverhttp.cpp
@@ -53,7 +53,7 @@ public:
 			auto it = wsConnectionMap.begin();
 			WebSocketOverHttp *sock = it->first;
 			wsConnectionMap.erase(it);
-			delete sock;
+			deleteSocket(sock);
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


src/proxy/websocketoverhttp.cpp
```diff
@@ -887,19 +880,14 @@ class WebSocketOverHttp::Private : public QObject
...
-			q->framesWritten(reqFrames, reqContentSize);
+			q->framesWritten(framesRemoved, contentRemoved);
 			if(self.expired())
 				return;
 		}
 
-		bool hadContent = reqContentSize > 0;
-
-		reqFrames = 0;
-		reqContentSize = 0;
```
jkarneges, line 900: For consistency, don't reset these here. These and other related vars already get reset the next time they matter, in `update()`. `reqContentSize` was being reset here because `writeBytesAvailable()` was depending on it, but this is no longer the case.

mhp (approved):

PR #48198, opened 2025-04-05
fs: add single file watcher using notify crate

jkarneges:
This adds `FileWatcher`, for detecting changes to a single file. It is intended for use by the proxy to monitor the routes file, which it currently does using `QFileSystemWatcher`.

`FileWatcher` uses the [`notify`](https://github.com/notify-rs/notify) crate under the hood, consuming raw events without a "debouncer" since the proxy already has its own such debouncing logic. In order to detect file creates and tolerate file replacements, the parent directory is monitored instead of the specified file, and events related to other files are ignored.

Events are communicated to the application via file descriptor notifications, for compatibility with `SocketNotifier` in C++. The application should never read from the file descriptor directly. It should only register it in a poller to get notifications. After receiving a notification, `FileWatcher::file_changed()` should be called to check for changes, which will perform a read internally. This style of API, handing the application a file descriptor for notifications only and not for I/O, is inspired by the ZeroMQ API.

### DIFF
--- a/src/core/fs.rs
+++ b/src/core/fs.rs
@@ -150,32 +150,32 @@ fn set_fd_nonblocking(fd: RawFd) -> Result<(), io::Error> {
     Ok(())
 }
 
-struct WatchState {
-    _watcher: notify::RecommendedWatcher,
+struct FileWatcherState {
+    watcher: notify::RecommendedWatcher,
     changed: bool,
 }
 
-struct WatchData {
+struct FileWatcherData {
     file: PathBuf,
     read_fd: RawFd,
     write_fd: RawFd,
-    state: Mutex<Option<WatchState>>,
+    state: Mutex<Option<FileWatcherState>>,
 }
 
-pub struct Watch {
-    data: Arc<WatchData>,
+pub struct FileWatcher {
+    data: Arc<FileWatcherData>,
 }
 
 #[derive(Debug)]
-pub struct WatchError;
+pub struct FileWatcherError;
 
-impl Watch {
-    pub fn new<P: AsRef<Path>>(file_path: P) -> Result<Self, WatchError> {
+impl FileWatcher {
+    pub fn new<P: AsRef<Path>>(file_path: P) -> Result<Self, FileWatcherError> {
         let file = file_path.as_ref();
 
         let dir = match file.parent() {
             Some(p) => p,
-            None => return Err(WatchError),
+            None => return Err(FileWatcherError),
         };
 
         let mut fds = [0; 2];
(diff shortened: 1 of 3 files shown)


src/core/fs.rs
```diff
@@ -117,3 +121,246 @@ pub fn set_group(path: &Path, group: &str) -> Result<(), io::Error> {
...
+        };
+
+        let mut fds = [0; 2];
+
+        // SAFETY: fds pointer is valid
+        let ret = unsafe { libc::pipe(fds.as_mut_ptr()) };
+        assert_eq!(ret, 0);
+
+        for fd in &fds {
+            set_fd_nonblocking(*fd).unwrap();
```
deg4uss3r, line 188: We return a `FileWatcherError` here, why do we not bubble this unwrap up (along with a few others around obtaining locks) and panic here instead, should this fail if we cannot set `nonblocking` on _any_ of the files?

src/core/fs.rs
```diff
@@ -117,3 +121,246 @@ pub fn set_group(path: &Path, group: &str) -> Result<(), io::Error> {
...
+                        return;
+                    }
+                };
+
+                if !event.paths.into_iter().any(|p| p == data.file) {
+                    // skip unrelated events
+                    return;
+                }
+
+                if let Some(state) = data.state.lock().unwrap().as_mut() {
```
deg4uss3r, line 215: Nitpick: state lock unwrap that should at least be an `.expect()` to help if this ever goes wrong.

src/core/fs.rs
```diff
@@ -117,3 +121,246 @@ pub fn set_group(path: &Path, group: &str) -> Result<(), io::Error> {
...
+                        };
+                        assert!(ret == 1 || get_errno() == libc::EAGAIN);
+                    }
+                }
+            })
+            .expect("failed to create file watcher")
+        };
+
+        {
+            let mut state = data.state.lock().unwrap();
```
deg4uss3r, line 235: Same Nitpick: state lock unwrap that should at least be an .expect() to help if this ever goes wrong.

src/core/fs.rs
```diff
@@ -117,3 +121,246 @@ pub fn set_group(path: &Path, group: &str) -> Result<(), io::Error> {
...
+            }
+        }
+
+        Ok(Self { data })
+    }
+
+    pub fn file_changed(&self) -> bool {
+        let mut changed = false;
+
+        if let Some(state) = self.data.state.lock().unwrap().as_mut() {
```
deg4uss3r, line 257: Same Nitpick: state lock unwrap that should at least be an .expect() to help if this ever goes wrong.

src/core/fs.rs
```diff
@@ -117,3 +121,246 @@ pub fn set_group(path: &Path, group: &str) -> Result<(), io::Error> {
...
+            state.changed = false;
+        }
+
+        changed
+    }
+}
+
+impl Drop for FileWatcher {
+    fn drop(&mut self) {
+        let mut state = self.data.state.lock().unwrap();
```
deg4uss3r, line 282: What happens if we can't get `state` here, does this need to panic or could we try a backoff and drop later? Leaving dangling FDs might put us in a bad state (see Notary 😅)

jkarneges: The only time `lock()` fails is if the mutex is poisoned, in which case a panic is warranted. The project aborts on panic (we use `panic = abort` in the cargo settings), so no chance of leaking FDs.

src/core/fs.rs
jkarneges: As discussed, this operation shouldn't fail and isn't really recoverable anyway.

deg4uss3r (approved):

PR #48209, opened 2025-04-16
proxy: add support for new event loop in engine and domainmap threads

jkarneges:
Following #48207, this makes the `new_event_loop=true` setting cause the new event loop to be used in the proxy's other threads too, not only the main thread.

I've done some basic testing of common flows (start/stop, passthrough, websocket-over-http, accept/retry) and things look ok so far.

### DIFF
--- a/src/handler/handlerapp.cpp
+++ b/src/handler/handlerapp.cpp
@@ -409,9 +409,9 @@ private:
 		{
 			log_debug("using new event loop");
 
-			// enough for control requests and prometheus requests. client
-			// sessions don't use socket notifiers
-			int socketNotifiersMax = SOCKETNOTIFIERS_PER_SIMPLEHTTPREQUEST * (CONTROL_CONNECTIONS_MAX + PROMETHEUS_CONNECTIONS_MAX);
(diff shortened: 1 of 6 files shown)


src/handler/handlerapp.cpp
```diff
@@ -409,9 +409,9 @@ class HandlerApp::Private
 		{
 			log_debug("using new event loop");
 
-			// enough for control requests and prometheus requests. client
-			// sessions don't use socket notifiers
-			int socketNotifiersMax = SOCKETNOTIFIERS_PER_SIMPLEHTTPREQUEST * (CONTROL_CONNECTIONS_MAX + PROMETHEUS_CONNECTIONS_MAX);
+			// enough for control requests and prometheus requests, plus an
+			// extra 100 for misc. client sessions don't use socket notifiers
```
mhp, line 413: > // extra 100 for misc.

😆

mhp (approved):
LGTM

src/handler/handlerapp.cpp
jkarneges: resource estimation is hard!

PR #48222, opened 2025-05-02
zhttpsocket: for client mode, add ability to receive via router socket

jkarneges:
Similar to #48073, this adds support for the "router response" feature to zhttpsocket's client mode. It enables receiving response messages via ROUTER, and indicating the source of the message in a flag (true if from the ROUTER socket, false if from the SUB socket). Existing handling code ignores the flag and assumes messages come from the SUB socket, so there is no change in behavior. In a future PR we can change the handling code to make use of the flag to properly handle router response messages.

### DIFF
--- a/src/connmgr/zhttpsocket.rs
+++ b/src/connmgr/zhttpsocket.rs
@@ -1,5 +1,6 @@
 /*
  * Copyright (C) 2020-2023 Fanout, Inc.
+ * Copyright (C) 2025 Fastly, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


src/connmgr/zhttpsocket.rs
```diff
@@ -2619,9 +2655,9 @@ impl AsyncClientStreamHandle {
         }
     }
 
-    pub async fn recv(&self) -> Result<arena::Arc<zmq::Message>, io::Error> {
+    pub async fn recv(&self) -> Result<(arena::Arc<zmq::Message>, bool), io::Error> {
         match self.receiver.recv().await {
-            Ok(msg) => Ok(msg),
```
serra-fastly, line 2624: Can use map_err here instead:
self.receiver.recv().await.map_err(|_|  io::Error::from(io::ErrorKind::BrokenPipe))

jkarneges: Ah yes, this works here since there is only one error kind.

deg4uss3r (approved):

PR #48227, opened 2025-05-15
runner: look for bins in target dir instead of bin dir

jkarneges:
Currently, the runner looks for component programs in a relative `bin` dir, to support running Pushpin within the source tree. However, the programs are placed there by `make` and not `cargo`, which means if you prefer developing using cargo you have to remember to run `make` before running Pushpin to ensure the latest bins actually get used.

This PR changes the runner to instead look in the `target/{buildmode}` dir for programs. That way, the runner will always use the latest programs without needing to run `make` first. This helps reduce the difference between using `make` and `cargo`, though it's still not perfect (see notes below).

With this change, the following build & run flows are possible from a clean tree:

* `cargo build` followed by `cargo run`
* `make` followed by `./pushpin`

Notes:

* `make` not only copies the component programs to the `bin` dir but it also copies the runner to the base dir as `./pushpin`. This means running `cargo build` followed by `./pushpin` will either run an old build from an earlier `make` invocation or fail due to the program not existing. Running `make` followed by `cargo run` is okay though.
* `cargo run` only builds the default executable before running it, so unfortunately that command alone is not enough to build & run Pushpin out of the box (since the other components won't have been built), and you also can't iterate with it (since it won't build the other components). Use `cargo build && cargo run` instead.

It would be great if someday `cargo build` could place the default executable in the base dir and `cargo run` could build all dependent executables, but as far as I can tell this is not possible to do yet.

### DIFF
--- a/src/core/mod.rs
+++ b/src/core/mod.rs
@@ -47,6 +47,10 @@ use std::os::unix::ffi::OsStrExt;
 #[cfg(test)]
 use std::path::{Path, PathBuf};
 
+pub fn is_debug_build() -> bool {
+    cfg!(debug_assertions)
+}
+
 pub fn version() -> &'static str {
     env!("APP_VERSION")
 }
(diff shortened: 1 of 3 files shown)


deg4uss3r:
Yeah unfortunately without some complex `build.rs` I don't think this is possible still :( 
https://internals.rust-lang.org/t/provide-a-way-to-always-build-required-binaries/19403

deg4uss3r (approved):

PR #48228, opened 2025-05-15
add "cow" types, initially as wrappers around qt types

jkarneges:
Long term we want to move off of Qt, and this will mean replacing all use of Qt types with either C++ standard library types or wrapped Rust types. For example, we could imagine replacing `QString` with `std::string`. These two types don't work exactly the same way though, and porting the entire codebase to types with different APIs and behaviors would be a lot of work as well as error prone. To simplify the porting process, I propose we substitute the Qt types with our own internal types that provide identical APIs and behaviors. Only the type names would be different. This should make the substitution process fairly safe.

To start out, this PR adds types ~~`AcString` and `AcByteArray`~~ `CowString` and `CowByteArray` as drop-in substitutes for `QString` and `QByteArray`, and uses them in `QZmq::Socket`. The ~~"Ac" stands for~~ "Cow" prefix means atomic reference count & copy on write, which is the main special behavior [provided by these (and most) Qt types](https://doc.qt.io/qt-6/implicit-sharing.html). Initially, the implementations are simply wrappers around the Qt types. For example, ~~`AcString`~~ `CowString` has a single private `QString` member that it forwards to. Later on, we could change it to wrap an ARC'd COW'd standard C++ or Rust string instead.

The substitute types are also cheaply convertible to/from Qt types, so they can be introduced gradually with minimal/zero overhead. For example, a `QString` can be converted to ~~an `AcString`~~ a `CowString` and back without deep copying. ~~The memory layouts of the types are also statically asserted as identical (similar to Rust's `#[repr(transparent)]`), to enable converting between collections without copying. For example, `QList<QByteArray> &` can be casted to `QList<AcByteArray> &`. It's a little scary looking but I *think* it is safe. Not sure if portable, but hopefully.~~

In order for collections of `Cow*` types to support cheap conversions to/from Qt collections types, as well as to support cheap conversions of individual elements, we can introduce non-templated collections types as needed. To start out, lists of `CowByteArray` can be managed using `CowByteArrayList`, internally backed by `QList<QByteArray>`.

Plan:

1. Provide substitute types with identical APIs that wrap Qt types, along with cheap conversions to/from Qt types. Gradually update the codebase to use these types.
2. Once the whole codebase is using the substitute types, remove any conversion functions to confirm elimination of direct use of Qt types.
3. Replace the implementations of the substitute types to be based on non-Qt types.
4. Remove the Qt library as a dependency of the project.


jkarneges:
> For example, `QList<QByteArray> &` can be casted to `QList<AcByteArray> &`. It's a little scary looking but I _think_ it is safe.

It seems this is not safe and crashes on Linux. Google says "don't do that", which is probably fair, though it would be nice to understand why. Even if it's simply against the spec, I'm curious what the compiler generates that would fail.

In any case, instead of trying to cast between template types I think we can explicitly add any needed list types, which I'm trying now with `AcByteArrayList` which wraps `QList<QByteArray>`. This enables the same kind of cheap conversion between outer and inner. However, in order to cheaply look up items it does rely on being able to cast between the item types (`QByteArray &` and `AcByteArray &`). This casting does not crash for me and is hopefully valid C++ since the types are not templates, but will double check that.

jkarneges:
> hopefully valid C++ since the types are not templates, but will double check that

Nope, not valid either. Something about violating strict aliasing.

I think I found a workaround by adding our own reference types. For example, instead of doing a questionable cast from `QByteArray &` to `AcByteArray &`, we can wrap the `QByteArray &` in an `AcByteArrayRef` with methods that forward to the inner ref.

jkarneges:
Ok I think this is finally workable. I've updated the description.

deg4uss3r (approved):
I don't see anything preventing us from using this!

PR #48231, opened 2025-05-28
Should chunked transfer has a size of a chunk?

sergey-shandar:


jkarneges:
Fair question. These examples show the body with the transfer encoding decoded, similar to the output of curl, which I think is reasonable for readability.

PR #48232, opened 2025-07-03
Ports command line argument handling for proxy and handler to Rust

serra-fastly:
This change is a part of the transition from C++ to Rust.

Argument handling has been collapsed into core/[handler/proxy]argsdata.cpp and core/[handler/proxy]cliargs.rs for C++ and Rust respectively. Old argument parsing logic was removed from the C++ modules. Runner was updated with the new command line values to standardize tags. ArgsData in Runner was updated to RunnerArgsData for differentiation.

In the Rust modules, handler and proxy use **clap** to parse arguments into CliArgs, convert to its FFI struct equivalent [Handler/Proxy]CliArgsFfi, and then call handler_main and proxy_main respectively with the [Handler/Proxy]CliArgsFfi variable being passed in as the argument. **cbindgen** creates a binding for [Handler/Proxy]CliArgsFfi within "rust/bindings.h" for use by C++ at compilation time. Then in the C++ modules, the [Handler/Proxy]CliArgsFfi gets loaded into an ArgsData struct for use with QT types and loaded into the Settings struct.

cbindgen was used to pass the data through because the linker had trouble with cxx due to QT (that wasn't able to be resolved with cxx-qt), but it'll be good to use cxx in the future for security and stability to create safe bindings (vs the unsafe bindings from cbindgen).

I've created unit testing for both the Rust modules that send the arguments within cliargs.rs and the C++ modules that receive the modules within [handler/proxy]argstest.cpp. Both test with the same values and the same two use cases— all arguments are set and no arguments are set.

Tested on a canary node (lck) on 08/08/2025.


deg4uss3r (commented):
PR looks good! Minus one debugging statement left in I think this is ready to test on a node for correctness.

deg4uss3r (approved):
Everything here looks good to me, going to tag in Justin though for a more thorough review of the c++ pieces.

src/core/ccliargs.rs
```diff
@@ -83,6 +83,87 @@ impl CCliArgs {
...
+    use std::ffi::CString;
+
+    #[repr(C)]
+    pub struct CCliArgsFfi {
+        config_file: *mut libc::c_char,
+        log_file: *mut libc::c_char,
+        log_level: libc::c_uint,
+        ipc_prefix: *mut libc::c_char,
+        port_offset: libc::c_uint,
+        routes: *mut libc::c_char, // Changed to single string instead of array
```
jkarneges, line 96: Good start. The `routes` field should be an array.

build.rs
```diff
@@ -353,7 +353,11 @@ fn get_qt_lib_prefix(lib_dir: &Path, version_maj: u16) -> Result<String, Box<dyn
 }
 
 fn find_boost_include_dir() -> Result<PathBuf, Box<dyn Error>> {
-    let paths = ["/usr/local/include", "/usr/include"];
+    let paths = [
```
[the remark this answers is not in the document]
serra-fastly: Merged!

src/core/ccliargs.rs
serra-fastly: Changed to array
https://github.com/fastly/pushpin/blob/8bfd1ddcc602210529bf59f934f5219838f2a823/src/core/ccliargs.rs#L93

src/core/ccliargs.rs
```diff
@@ -0,0 +1,394 @@
...
+        log_file: *mut libc::c_char,
+        log_level: libc::c_uint,
+        ipc_prefix: *mut libc::c_char,
+        port_offset: libc::c_int,
+        routes: *mut *mut libc::c_char,
+        routes_count: libc::c_uint,
+        quiet_check: libc::c_int,
+    }
+
+    impl CCliArgsFfi {
```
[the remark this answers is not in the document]
serra-fastly: 5142efb265e412a9081d36102563c0b1d21bfed9

src/core/ccliargs.rs
```diff
@@ -0,0 +1,394 @@
...
+    pub fn to_ffi(&self) -> ffi::CCliArgsFfi {
+        ffi::c_cli_args_to_ffi(self)
+    }
+}
+
+pub mod ffi {
+    use std::ffi::CString;
+
+    #[repr(C)]
+    pub struct CCliArgsFfi {
```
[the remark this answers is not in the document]
serra-fastly: ad3cd4949f90edef7b0849fb8907d485d8e177fc

log.txt
jkarneges, line 1: accidental add I assume

src/core/handlercliargs.rs
jkarneges, line 1: Let's move this into the handler module.

src/core/proxycliargs.rs
jkarneges, line 1: And this into the proxy module.

Cargo.toml
```diff
@@ -82,7 +83,7 @@ bench = false
 
 [[bin]]
 name = "pushpin-handler"
-test = false
+test = true
```
jkarneges, line 86: is this getting used?

src/core/proxycliargs.rs
```diff
@@ -0,0 +1,337 @@
...
+        pub log_level: libc::c_uint,
+        pub ipc_prefix: *mut libc::c_char,
+        pub routes: *mut *mut libc::c_char,
+        pub routes_count: libc::c_uint,
+        pub quiet_check: libc::c_int,
+    }
+
+    // Converts CliArgs to a C++-compatible struct
+    #[no_mangle]
+    pub extern "C" fn proxy_cli_args_to_ffi(args: &super::CliArgs) -> ProxyCliArgsFfi {
```
jkarneges, line 95: This function is only called by Rust (and also only usable by Rust due to the `&`) so it can be moved out of the ffi module.

src/core/proxycliargs.rs
```diff
@@ -0,0 +1,337 @@
...
+    ///
+    /// This function is unsafe because it takes ownership of raw pointers and frees their memory.
+    /// The caller must ensure that:
+    /// - The `ffi_args` struct was created by `proxy_cli_args_to_ffi`
+    /// - Each pointer field in `ffi_args` is either null or points to valid memory allocated by `CString::into_raw()` or `libc::malloc()`
+    /// - The `routes` array and its individual string elements were allocated properly
+    /// - No pointer in `ffi_args` is used after this function is called (double-free protection)
+    /// - This function is called exactly once per `ProxyCliArgsFfi` instance
+    #[no_mangle]
+    pub unsafe extern "C" fn destroy_proxy_cli_args(ffi_args: ProxyCliArgsFfi) {
```
jkarneges, line 176: This function also only Rust-usable.

src/core/settings.h
```diff
@@ -25,22 +25,25 @@
...
 
 class QSettings;
 
 class Settings
 {
 public:
+
 	Settings(const QString &fileName);
 	~Settings();
-
```
jkarneges, line 36: not sure about this whitespace change

src/core/settings.h
```diff
@@ -25,22 +25,25 @@
 
 #include <QString>
 #include <QVariant>
+#include "rust/bindings.h"
```
jkarneges, line 28: is this needed?

src/core/tests.pri
```diff
@@ -6,3 +6,4 @@ SOURCES += \
 	$$PWD/tcpstreamtest.cpp \
 	$$PWD/unixstreamtest.cpp \
 	$$PWD/eventlooptest.cpp
+
```
jkarneges, line 9: whitespace

src/handler/handlerapp.h
```diff
@@ -23,14 +23,15 @@
 
 #ifndef HANDLERAPP_H
 #define HANDLERAPP_H
+#include "settings.h"
```
jkarneges, line 26: and maybe this only needs to be "rust/bindings.h" ?

src/handler/handlerargstest.cpp
```diff
@@ -0,0 +1,99 @@
...
+
+void handlerargstest()
+{
+
+    // Create dummy argc/argv for QCoreApplication
+    int argc = 1;
+    char appName[] = "pushpin-handler";
+    char* argv[] = { appName, nullptr };
+
+    QCoreApplication qapp(argc, argv);
```
jkarneges, line 39: we are trying to avoid `QCoreApplication` now in tests. I wonder if this can be removed and it still works?

src/core/proxycliargs.rs
```diff
@@ -0,0 +1,337 @@
...
+    about = "Pushpin proxy component."
+)]
+pub struct CliArgs {
+    /// Set path to the configuration file
+    #[arg(short, long, value_name = "file")]
+    pub config_file: Option<String>,
+
+    /// Set path to the log file
+    #[arg(short = 'l', long, value_name = "file")]
+    pub log_file: Option<String>,
```
jkarneges, line 37: Let's retain the existing argument styles, e.g. `--logfile` instead of `--log-file`. That way we don't break existing usage and also the runner doesn't need to be updated.

For v2 I do want to update the styles which we can plan to do around the time of such a release.

src/core/proxycliargs.rs
serra-fastly: bb5eb4d6de3ef8f63640ee30e9fc979c38d59c1b

log.txt
serra-fastly: c0810cec0ac17382972c4eddabf45b5347f9a312 sorry yes I kept manually deleting it, didn't realize it made its way out

src/handler/handlerapp.h
serra-fastly: f5b2f5deacf8136de02a2e20140bd7edb4881b80

src/core/proxycliargs.rs
serra-fastly: got it okay, I was thinking I could start updating it now but heard d45330da5a04221dec2933922182622167b01e6b

src/core/proxycliargs.rs
serra-fastly: dbf51be1e2b0ec9eec2d7f6bedb24a8ca1752b52

src/runner/pushpinproxyservice.cpp
```diff
@@ -52,7 +52,7 @@ PushpinProxyService::PushpinProxyService(
 		args_ += "--loglevel=" + QString::number(logLevel);
 
 	foreach(const QString &route, routeLines)
-		args_ += "--route=" + route;
+		args_ += "--routes=" + route;
```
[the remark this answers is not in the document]
serra-fastly: 09d4edc19b09556a22ac1b68b2f790d9354d93cb sorry my mistake

src/runner/service.rs
```diff
@@ -322,21 +322,26 @@ impl PushpinProxyService {
...
             Some(&x) => x,
             None => settings.log_levels.get("default").unwrap().to_owned(),
         };
         args.push(format!("--loglevel={}", log_level));
 
-        for route in settings.route_lines.clone() {
-            args.push(format!("--route={}", route));
-        }
+        // Routes
+        args.push(format!("--routes={}", settings.route_lines.join(",")));
```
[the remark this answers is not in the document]
serra-fastly: 09d4edc19b09556a22ac1b68b2f790d9354d93cb

.gitignore
```diff
@@ -15,6 +15,7 @@ target_wrapper.sh
 vendor
 .vscode
 conf.pri
+log.txt
```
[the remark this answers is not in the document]
serra-fastly: updated, moved log setup into run and out of argdata, made test logfile names unique just in case for the future f32d28f2657fa6edcdc1c62ddd6a8cc59d090d70

src/handler/handlerapp.cpp
```diff
@@ -98,6 +98,34 @@ class HandlerApp::Private
...
+		else
+			log_setOutputLevel(LOG_LEVEL_INFO);
+
+		// Set the log file if specified
+		if(!args.logFile.isEmpty())
+		{
+			if(!log_setFile(args.logFile))
+			{
+				log_error("failed to open log file: %s", qPrintable(args.logFile));
+				exit(1);
```
jkarneges, line 113: This can become `return 1` now that the logic is back where it was.

src/handler/handlerapp.cpp
```diff
@@ -98,6 +98,34 @@ class HandlerApp::Private
...
+
+		log_debug("starting...");
+
+		// QSettings doesn't inform us if the config file can't be opened, so do that ourselves
+		{
+			QFile file(args.configFile);
+			if(!file.open(QIODevice::ReadOnly))
+			{
+				log_error("failed to open %s", qPrintable(args.configFile));
+			}
```
jkarneges, line 126: need `return 1` here too

jkarneges (approved):

PR #48243, opened 2025-08-07
fix warnings on latest rust

jkarneges:
Updated the `fanout/build-base` image with the latest Rust, which the jobs on this PR ran against.


clintjedwards (approved):
Long live 1.89

PR #48252, opened 2025-08-19
cowbytearray: add more methods

jkarneges:
This adds a bunch of methods to `CowByteArray` that we'll need in order to use it in more places. Things like `indexOf`, `mid`,`operator[]`, iterators, etc.


clintjedwards (approved):
This reveals how bad I am at c++, but I've never seen the `friend` keyword before 😂. Very cute naming c++ maintainers.

PR #48328, opened 2026-04-23
reformat all c++ files with clang-format

jkarneges:
We should autoformat the C++ code similar to how we use rustfmt for the Rust code. This PR uses `clang-format` to reformat all of the C++ code to a consistent, modern style. I did not attempt to preserve the existing style which was doing some things I think we should phase out, such as braces on new lines. This change only contains the reformatting and nothing else. Later, we can add a `.clang-format` file, update CI to enforce style, etc.

To review, I suggest performing the same mass reformatting locally and confirming the resulting diff is the same as this one. To do that:

* Install `clang-format`

* Add the following `.clang-format` file to the root of the project:

```
Language: Cpp
BasedOnStyle: LLVM

# Use 4 spaces to match rustfmt
IndentWidth: 4
ContinuationIndentWidth: 4
ConstructorInitializerIndentWidth: 4

# Keep braces on the same line (Rust/Modern C++ style)
BreakBeforeBraces: Attach

# Clean layout
ConstructorInitializerAllOnOneLineOrOnePerLine: true
AccessModifierOffset: -4
ColumnLimit: 100
```

* Run `find src -name "*.cpp" -o -name "*.h" | xargs clang-format -i` from the project root. As far as I can tell, clang-format can't crawl a project tree on its own so we have to use something like `find` to hit every file.

### DIFF
--- a/src/core/bufferlist.cpp
+++ b/src/core/bufferlist.cpp
@@ -22,164 +22,144 @@
 
 #include <assert.h>
 
-BufferList::BufferList() :
-	size_(0),
-	offset_(0)
-{
+BufferList::BufferList() : size_(0), offset_(0) {}
+
+void BufferList::findPos(int pos, int *bufferIndex, int *offset) const {
+    assert(pos < size_);
+
+    int at = 0;
+    int curOffset = offset_;
+
+    while (true) {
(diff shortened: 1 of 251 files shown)


deg4uss3r (approved):
yeah... LGTM

PR #48335, opened 2026-04-29
filewatcher: support relative paths

jkarneges:
I was recently messing around with `DomainMap` and the routes file watching wasn't working, and after tearing up the carpet it turned out it was because I used a relative path as input. This was breaking the underlying `FileWatcher`'s path comparisons. Currently we always provide an absolute path to `DomainMap`, but I figure it would be good to update `FileWatcher` to support relative paths to avoid such surprises. This PR does this by internally canonicalizing the parent path.

### DIFF
--- a/src/core/fs.rs
+++ b/src/core/fs.rs
@@ -182,8 +182,8 @@ pub struct FileWatcher {
 
 #[derive(Debug, Error)]
 pub enum FileWatcherError {
-    #[error("no parent")]
-    NoParent,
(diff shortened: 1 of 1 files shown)


clintjedwards (approved):
If I had a nickel for the amount of times I wrote something with file paths and then had to go back to fix a really confusing error because of relative paths; 

I wouldn't be rich, but I would have a lot of nickels.

PR #48345, opened 2026-05-09
list: fix circular reference leaks in rc tests

jkarneges:
Unlike slab-based lists, dropping Rc-based lists doesn't necessarily drop the nodes since the nodes maintain strong, bidirectional references to each other. The simple fix is to remove each individual node from such lists before dropping the list. We aren't using Rc-based lists anywhere yet so the leaks are only in these tests.


clintjedwards (approved):
Could refactor into a `clear` method, but this is fine.

jkarneges:
Yeah I'm with you on that. Will follow up with something.

PR #48347, opened 2026-05-13
use maybe_unused attribute instead of Q_UNUSED macro

jkarneges:
`maybe_unused` is available as of C++17.

### DIFF
--- a/src/handler/handlerengine.cpp
+++ b/src/handler/handlerengine.cpp
@@ -2080,11 +2080,7 @@ private:
     }
 
 private:
-    void report_finished(const DeferredResult &result) {
-        Q_UNUSED(result);
-
-        report.reset();
-    }
+    void report_finished([[maybe_unused]] const DeferredResult &result) { report.reset(); }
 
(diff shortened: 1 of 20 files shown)


src/handler/handlerengine.cpp
```diff
@@ -2080,11 +2080,7 @@ class HandlerEngine::Private {
     }
 
 private:
-    void report_finished(const DeferredResult &result) {
-        Q_UNUSED(result);
-
-        report.reset();
-    }
+    void report_finished([[maybe_unused]] const DeferredResult &result) { report.reset(); }
```
serra-fastly, line 2083: should this be still in three lines, just to keep the formatting consistent?

jkarneges: It's a little weird, but our current clang-format config shrinks down small functions like that.

serra-fastly: Yeah I had a feeling it was the autoformatter but still wanted to check. Sounds good!

serra-fastly (approved):

PR #48358, opened 2026-06-02
refactor: Consolidate docker into main pushpin repo

clintjedwards:
In an effort to simplify the release processes we have I think it would be good to consolidate [the docker repo](https://github.com/fanout/docker-pushpin) into this main repo. To my knowledge, there is no real advantage to having a separate repo for this functionality (with the exception of maybe Github permissioning) so consolidating it here and running the CI as part of our normal release process should cut down on some steps.

The only change I've made from the repo is modifying the "MAINTAINER" label into `LABEL maintainer =` since the MAINTAINER label is deprecated. And I also added Fastly OSS as a maintainer.

The CI job will be in a follow up PR.

PR #48361, opened 2026-06-03
refactor: Consolidate pushpin-debian repo in to main pushpin repo

clintjedwards:
Following on the heels of the docker consolidation (https://github.com/fastly/pushpin/pull/48358) I think it makes sense to also consolidate the debian package creation to simplify and unify the release process. We've basically moved most of the files under `packaging/debian` and merged/removed some of the others to facilitate running `cargo-deb` from root.

Some of the changes made:

What moved:

* All debian packaging files (man pages, changelog, NEWS, maintainer scripts, pushpin.init, pushpin.logrotate, pushpin.tmpfile, prepare-dist.sh) now live in packaging/debian/

What changed:

* `[package.metadata.deb]` is now in the root Cargo.toml instead of a separate one. This means the version is always in sync with the main package, no separate bump needed.
* `cargo deb` is now run from the repo root with no --manifest-path flag
* `prepare-dist.sh` now uses git worktree against the local repo instead of cloning from GitHub; it also no longer takes a version argument (just the git ref)
* dch -c now points to packaging/debian/changelog
* Distro-specific branches (fanout-noble, etc) will live in the main repo going forward

What's the same:
* The overall release workflow is mostly unchanged. (checkout the distro branch, run prepare-dist.sh, update changelog, run cargo deb)


clintjedwards:
I'll also have to recreate the distro branches and test this on the next release.

clintjedwards:
> I'll also have to recreate the distro branches and test this on the next release.

Justin pointed out that if we also have to manage the branches it might be better to do that in a different repo, but when investigating the only thing we really use the branches for is changelog management, the biggest changes being the appended headers: `pushpin (1.41.0-1~noble1) fanout-noble; urgency=medium`.

This is technically convention, but does not affect `apt changelog <package>` output and probably wont matter to the vast majority of people. We can just use the single changelog and that should work fine (and simpler).

I've updated the PR to reflect this new process, dropping the distro branches

jkarneges (approved):

PR #48366, opened 2026-06-13
buffer: add helper function for implementing Write::write_vectored

jkarneges:
The default implementation of [Write::write_vectored](https://doc.rust-lang.org/std/io/trait.Write.html#method.write_vectored) conservatively calls `write` with the first slice only. This can lead to surprising behavior when developing code that performs vectored writes against memory/fake destinations that don't implement `write_vectored` themselves, and only the first slice shows up.

This PR adds a function that calls `write` on all provided slices, which can be a more convenient default, at least for in-memory destinations where the call costs are cheap. It then uses it in a couple of our buffer types.

### DIFF
--- a/src/core/buffer.rs
+++ b/src/core/buffer.rs
@@ -448,6 +448,38 @@ impl<'a: 'b, 'b> LimitBufsMut<'a, 'b> for [&'a mut [u8]] {
     }
 }
 
+pub fn write_trait_vectored_helper<W: Write>(
+    w: &mut W,
+    bufs: &[io::IoSlice],
+) -> Result<usize, io::Error> {
+    // Like std, if bufs is empty then we'll call write with no data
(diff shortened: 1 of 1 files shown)


clintjedwards (approved):

jkarneges:
Tweaked the implementation a little bit. Previously, if an error is encountered, partial progress would be reported only when the error kind is `WriteZero`. However, I think we should always report partial progress when an error is encountered, regardless of type (with the exception of `Interrupted`, for which retrying is idiomatic), otherwise that information is lost. Of course, whenever we report partial progress in response to an error then we are losing the error information instead, but in general the next write attempt should trigger the same error again.

Note that these distinctions aren't meaningful when used directly by memory buffers, which is all we are doing today. Just making the impl a little more robust in case it finds its way into a middleware atop a socket or something.

PR #48370, opened 2026-06-18
track: add into_inner

jkarneges:
The `track` module is used to enforce lifetimes of IPC messages, such that only one such tracked message exists at a time per task and that the message is processed timely. For the upcoming work to send large response headers, we'll need to keep the backing message around longer than usual so it can be borrowed while its contents are being sent to the client. This PR adds an `into_inner` method to `Track`, making it possible for a task to optionally exclude a message from tracking.

Note: we didn't need this for sending large request headers to the server, since the source message for outbound requests already begins in an untracked state.

### DIFF
--- a/src/connmgr/track.rs
+++ b/src/connmgr/track.rs
@@ -55,6 +55,13 @@ impl<'a, T> Track<'a, T> {
             inner: Some(TrackInner { value, active }),
         }
     }
+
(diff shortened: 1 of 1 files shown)


jkarneges:
CI failing due to the code not being used yet. Will just put it in the PR that uses it.

PR #48374, opened 2026-06-22
update task sizes commit ids

jkarneges:


clintjedwards (approved):

jkarneges:
It occurs to me `git blame` might be sufficient for keeping track of which commit the numbers were updated at, without having to literally write the commit ID in the source file. Will consider that for next time. The current approach was more convenient back when we weren't squashing and didn't require follow-up commits.

clintjedwards:
> It occurs to me `git blame` might be sufficient for keeping track of which commit the numbers were updated at, without having to literally write the commit ID in the source file. Will consider that for next time. The current approach was more convenient back when we weren't squashing and didn't require follow-up commits.

I also wonder, since its part of a test, if you could automate it by writing it into the dev build step somehow. Feels like it should be doable and would cut down on having to actually remember to bump it and update the hash every time.

Then you could probably write another test just to check to make sure that those numbers don't inflate too far past some reasonable limit.

jkarneges:
The tricky part would be having both the old and new numbers available at the same time to do the comparisons. Maybe the threshold test could use git commands to fish out the previous numbers.

clintjedwards:
> The tricky part would be having both the old and new numbers available at the same time to do the comparisons. Maybe the threshold test could use git commands to fish out the previous numbers.

That would work. In theory you could also just generate a file almost like a 'golden' file with the previous numbers. Or write them into a comment above each number (if you're autogenerating the numbers anyway autogenerating the comment for them should be reasonable). If the goal is just comparing how much it bloated.

You could also just set a hard limit and then check in when past that limit. I assume we can do some back of the envelope math to figure out what size limits we're comfortable with roughly. Then if that bar is too low for the next change, the author just bumps the tests and that makes it so the new limits have to be reviewed anyway.

That threshold test would also provide a good place to put some in-code documentation around the significance of the numbers.

PR #48377, opened 2026-06-25
add ffi around rust url crate, with implicit sharing

jkarneges:
This adds an FFI around the `url` crate, as the basis for a `QUrl` replacement.

Notably, `QUrl` is an [implicitly shared](https://doc.qt.io/qt-6/implicit-sharing.html) type. To support similar behavior here, the data is wrapped in `Arc`, and `Arc::make_mut` is used for copy-on-write operations. This is done generically, by introducing a type `CowArc<T>` which implements copy-on-write behavior for any `T` via the `DerefMut` trait, and then declaring `CowUrl` as a newtype of `CowArc<Url>`.

The Rust `CowUrl` API is intentionally lean and excludes any `QUrl`-isms. My sense is there should only be just enough here to facilitate efficient cross-language usage, which basically means agreeing on a core type (`url::Url`) and a smart pointer (`Arc`), and then making an FFI for only those things. Everything else can live outside of the FFI as language-specific conveniences. In practice this will mean the Rust code can use `CowUrl`  more-or-less as if it were a `Url` thanks to auto deref, and the C++ code can have its own wrapper API designed to mimic `QUrl` along with special behavior such as a null state.

I should also note a limitation: the Rust `Url` type does not support relative URLs, whereas `QUrl` does. Since there are only a few places in the C++ code where we need relative URLs, and those URLs are always resolved against an absolute base URL before being used for anything, my plan is for the C++ code to simply store relative URLs in string types.

### DIFF
--- a/cbindgen.toml
+++ b/cbindgen.toml
@@ -9,4 +9,4 @@ after_includes = """
 
 [parse]
 parse_deps = true
-include = ["jsonwebtoken", "zmq"]
+include = ["jsonwebtoken", "url", "zmq"]
--- a/src/core/cow.rs
+++ b/src/core/cow.rs
@@ -1,0 +1,54 @@
+/*
+ * Copyright (C) 2026 Fastly, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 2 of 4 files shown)


clintjedwards (approved):
I had to look up a lot of how QUrl works to be able to match the logic choices, but looks great to me.

PR #48386, opened 2026-07-15
treat route log_level as the max level rather than the logging level

jkarneges:
Currently, the `log_level` route parameter controls the log level used by some route-specific log messages. This is mainly useful for enabling logs for certain routes without having to increase the application-wide max log level. However, this approach of changing the levels of log messages has some problems/limitations:

* The original levels are lost. For example, if the param is set to 1 to cause a route's info/debug/trace log calls to log at warning level, well, they all get logged as warnings making their individual severities unclear.
* It adds noise to the more critical levels. For example, causing a route's info/debug/trace log calls to log at warning level means that some warnings are real warnings and others might just be hoisted info/debug/trace messages.
* It's not possible to granularly control which route-specific log messages are targeted. For example, it could be nice to be able to enable info-level logs for a route without also enabling debug/trace levels.

This PR attempts to make things better by changing the behavior of `log_level` from controlling the levels of the messages themselves to controlling the max level allowed for such messages. For example, if the param is set to 2, then the route's info level logs will be logged regardless of the application-wide max level, and they will be logged as info level. Setting to 3 would then get the debug level logs too, and so on. If unspecified, the application-wide max level applies.

Below is an example of Pushpin running with a default max level of 1 and configured with two routes, with requests for the second route using a max level of 3:

```
% ./pushpin-legacy -m --route "* test" --route "example.com,log_level=3 test" --loglevel 1   
```

Requests to both routes work as expected:

```
% curl http://localhost:7999                           
Hello from the Pushpin test handler!
% curl --connect-to ::localhost:7999 http://example.com
Hello from the Pushpin test handler!
```

But only requests to the second one are logged:

```
[INFO] 2026-07-16 11:22:06.050 [proxy] GET http://example.com/ -> test code=200 37
```

The one caveat of this approach is it may be confusing to see higher level logs being outputted when the application-wide max level is set to a lower level. However, I think this should be less confusing the current behavior. The way to reason about it: the max level is not a fixed effect controlled by a single configuration option but a dynamic effect controlled by multiple configuration options. The application-wide max level is a default/fallback that may be superseded by context-specific max levels.

Compatibility note: in general we shouldn't change the behavior of existing options so as to not break Pushpin's semver guarantees, however I think we can get away with changing the behavior of the `log_level` route param because log output is not part of the guarantee. The worst that happens to existing users of the param is route-specific logs stop appearing until they update their configuration, but we are not obligated to retain such logs and could remove them too.

### DIFF
--- a/src/core/log.rs
+++ b/src/core/log.rs
@@ -585,7 +585,7 @@ impl Log for DebugLogger {
     fn flush(&self) {}
 }
 
-mod ffi {
+pub mod ffi {
     use super::*;
     use std::ffi::{c_char, c_int, CStr};
     use std::fs::OpenOptions;
@@ -606,7 +606,7 @@ mod ffi {
         }
     }
 
-    fn int_to_optional_level_filter(level: c_int) -> Option<LevelFilter> {
+    pub fn int_to_optional_level_filter(level: c_int) -> Option<LevelFilter> {
         match level {
             c_int::MIN..=-1 => None,
             0 => Some(log::LevelFilter::Error),
--- a/src/core/logutil.cpp
+++ b/src/core/logutil.cpp
@@ -64,9 +64,10 @@ static QString makeLastIdsStr(const HttpHeaders &headers) {
     return out;
 }
 
-static void logPacket(int level, const QString &message, const Variant &data = Variant(),
-                      int dataMax = -1, const QByteArray &content = QByteArray(),
-                      int contentMax = -1) {
+static void logPacketWithOutputLevel(int outputLevel, int level, const QString &message,
+                                     const Variant &data = Variant(), int dataMax = -1,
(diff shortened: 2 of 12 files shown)


clintjedwards (approved):
You could signify that a particular route has overridden log_level settings by doing something like:

* At the start of the service print some info log line signifying the route and the new/superseding log level.
* For each log line that is overridden you could append some "overridden by config" message or similar.

I'm more partial to the first one since it is probably easier to code and probably easier on the logging infra since it wont produce that many more characters than current.

Might help a user that isn't familiar with the behavior understand why certain logs are breaking the global level.

jkarneges:
Good suggestions. I like the first except that such an info-level log line would likely be filtered out by any setup that needs to use the route param. So, maybe the second.

PR #48391, opened 2026-07-28
consolidate tcp/unix socket listening configuration

jkarneges:
While preparing to implement prometheus metrics in connmgr, I realized we need a way to specify a listener binding, and currently we have two existing code paths for this kind of thing (for accepting client connections and for the debug socket) that are not reusable. This PR consolidates both the input parsing and the socket binding to reduce code duplication as well as make it reusable.

Notably, we need to support two kinds of inputs: strings that can specify either a tcp or unix listening configuration (used for accepting client connections, where a `local` param indicates unix), and strings for specifying only unix listening configuration (used for the debug socket, no `'local` flag needed in that context).

Summary:

* Input string parsing is moved into the `config` module. It knows how to parse various string forms, such as a tcp port by itself (`9000`), tcp addr+port (`127.0.0.1:9000`), or unix path with params (`/path/to/foo.sock,mode=700`) into a `NetListenConfig` struct.
* Functions are added to the `net` module for binding TCP or Unix sockets based on a supplied config.
* Call sites updated to use the new APIs.

### DIFF
--- a/src/bin/pushpin-connmgr.rs
+++ b/src/bin/pushpin-connmgr.rs
@@ -17,7 +17,8 @@
 
 use clap::{Arg, ArgAction, Command};
 use log::{error, LevelFilter};
-use pushpin::connmgr::{run, App, Config, DebugSocketConfig, ListenConfig, ListenSpec};
+use pushpin::connmgr::{run, App, Config, ListenConfig, ListenSpec};
+use pushpin::core::config::{NetListenConfig, UnixListenConfig};
 use pushpin::core::log::{get_simple_logger, local_offset_check};
 use pushpin::core::version;
 use std::error::Error;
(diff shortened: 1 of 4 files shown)


src/bin/pushpin-connmgr.rs
```diff
@@ -114,76 +115,47 @@ fn process_args_and_run(args: Args) -> Result<(), Box<dyn Error>> {
...
+        let (spec, stream) = match net {
+            NetListenConfig::Tcp(mut c) => {
+                let stream =
+                    c.params.remove("req").is_none() || c.params.remove("stream").is_some();
+                let tls = c.params.remove("tls").is_some();
+                let default_cert = c.params.remove("default-cert");
+
+                if let Some(k) = c.params.keys().next() {
+                    return Err(format!("failed to parse listen: invalid param: {}", k).into());
+                }
```
clintjedwards: This is a bit hard to evaluate without an example of the type of string we're targeting/parsing, but I see that we do an `||` whos first instance can in theory return true. Two things I'm not 100% sure about, but might be an issue:

1. If the evaluation is lazy then the or could return true at the first instance and then you never actually remove "stream" if its exists. I think they're opposite types so its probably pretty common for req to not be there if stream exists?
2. If we leave stream in the params hashmap, then when we check if there are any other keys on line 129 we'll always get an error right?

I haven't tested this assumption though. Should be fairly easy to write a test for.

clintjedwards (approved):
Looks good with one concern on the `.remove` statements.

src/bin/pushpin-connmgr.rs
jkarneges: Good catch. Short circuiting is a bad idea here. Best to extract all known params first, then evaluate. I've reworked things.

PR #48395, opened 2026-08-04
Bump version to 1.43.0

clintjedwards:
We're trying out a new(to this project) single trunk approach for releases and hoping to make them more frequent as we've made strides on fixing other parts of the release process.

This is the first step where we cut the release at a specific commit and bump the version.

Note: We're moving away from appending `-dev` since every version bump will essentially be a chronological marker for the exact commits within those versions.

Manually built binaries can be designated to have special suffixes by using specific build flags instead.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "pushpin"
-version = "1.42.0-dev"
+version = "1.43.0"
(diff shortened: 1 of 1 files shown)


clintjedwards:
Was not aware we actually mix v2 commits into main also. Need to rethink how to single branch from that position.

PR #48401, opened 2026-08-18
reformat all c++ files with clang-format

jkarneges:
This is simply `make fmt` in the v1 branch, after copying over the `.clang-format` file and Makefile directives from main.

We could get similar changes by merging main into v1, but I'm attempting to make main and v1 more diffable to each other without actually bringing in the changes that have occurred in main since the last release.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,15 @@
+Language: Cpp
+BasedOnStyle: LLVM
+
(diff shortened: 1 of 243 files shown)


jkarneges:
Closing as we no longer need to maintain a v1 branch for releases.

PR #48416, opened 2026-09-01
[ci] Run codeql only on files that require it

clintjedwards:
This change is to fix an earlier change which occurred for the same reasons.

We previously tried to make it so files like `.md` or `Dockerfile` wouldn't
trigger the full suite of tests that builds pushpin and tests it, purely
for quality of life reasons.

Unfortunately, this conflicts with the expected Github status checks settings
where Github expects certain checks to return before a PR can be merged.

To keep the status checks requirements intact but also allow us to skip
checks which don't make sense for the PR we have to move away from the
`paths-ignore` way of filtering to running a small pre-check workflow to
list the files being changed and return a value that the rest of the
workflow can read to determine if we should run or skip the rest of the
jobs.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -4,12 +4,31 @@ on:
   push:
     branches: [ "main" ]
   pull_request:
-    branches: [ "main" ]
   schedule:
     - cron: '21 13 * * 6'
 
 jobs:
+  changes:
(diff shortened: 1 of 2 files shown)


jkarneges (approved):

jkarneges (approved):
thanks I hate it

Issue #48357, opened 2026-06-02
Current Docker Ubuntu version (24.10) is end of life


clintjedwards:
Currently we use `24.10` for our Docker Ubuntu version which went EOL in July 2025. We should just update to whatever the latest (supported) LTS version is.

PR #48418, opened 2026-09-08
update ubuntu to 26.04 and fix docker build for v1.42.0

b3n3d1k7:
Updated ubuntu to 26.04 and changed the make command order to fix the docker build for v1.42.0

Issue: https://github.com/fastly/pushpin/issues/48357

I could also include the changes from https://github.com/fastly/pushpin/pull/48415 if that makes sense.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -3,7 +3,7 @@
 #
 
 # Pull the base image
-FROM ubuntu:24.10 as build
+FROM ubuntu:26.04 AS build
 
 # Install deps
 RUN \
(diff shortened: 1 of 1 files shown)


jkarneges:
The command reordering is a bit unfortunate. We should consider dropping `--all-features` from `make check` so that a conventional build->test->install flow works.

I should note that the `legacy-runner` feature was recently removed from `main`, so an install following a check might actually work now, but reordering is still the right thing to do to defend against future features breaking that.

clintjedwards (approved):

